PR #7, opened 2018-10-10
Added Obed_Metemey directories

ometemey:
This is the message for my completed assignments.

### DIFF
--- a/examples/UWPCE_Python/git_push_test_class1
+++ b/examples/UWPCE_Python/git_push_test_class1
@@ -1,0 +1,4 @@
+
+###This is a test push for git for the UW PCE Python Programming class###
+
+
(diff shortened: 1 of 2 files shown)


examples/UWPCE_Python/git_push_test_class1
```diff
@@ -0,0 +1,4 @@
+
```
PythonCHB: you shouldn't be adding anything to the examples dir.

you can remove this with:

git rm examples/UWPCE_Python/git_push_test_class1

then:

git commit -a

git push

and this PR will be updated.

PythonCHB:
Not merging 'till the extra file is removed

PythonCHB:
nice job -- always harder to figure out how to undo something...

PR #28, opened 2018-10-11
Python Pushups: exceptions file exercise

rushinstuffin:
Not sure if we were supposed to push/pull this, but here it is anyway. I think I messed up the commit here. I only wanted to commit the latest one. I'll get it right next time, I hope it's not a problem!


PythonCHB:
This is perfect -- the PR looks good, as does the code -- merging away :-)

PR #29, opened 2018-10-12
Added first exercise (break_me.py)

ktbzzz:


PythonCHB:
Nice work!

I wasn't expecting you to catch the Exceptions -- we'll get to that later in the class, but well done!

You will want to remove the space from your dir name -- all file systems, and git can handle spaces, but when doing anything on the command line, it can make it hard, and some tools do break.

`git mv "Jae Kim" JaeKim`

(or Jae_Kim)

Should do it. you need to quotes, so that the the space doesn't confuse git...

-CHB

PR #36, opened 2018-10-14
Added the break_me file in session01 dir for my branch

framllat:


students/student_framllat/session02/fizzbuzz.py
```diff
@@ -0,0 +1,15 @@
+import random
+
+for i in range(1,10):
+
+    test_num = random.randint(1,100)
```
PythonCHB: interesting -- why randomize? I think the problem description simple says the numbers from one to 100

students/student_framllat/session02/fizzbuzz.py
```diff
@@ -0,0 +1,15 @@
...
+#    print("This is the num:  ",test_num)
+
+    if (test_num % 3 == 0) and (test_num % 5 == 0):
```
PythonCHB: this is a good, solid solution.

but now that you are ahead of the game, can you figure out how not to check for `% 3` and `% 5` twice each?

students/student_framllat/session01/break_me.py
```diff
@@ -0,0 +1,70 @@
...
+def error_name():
+    while True:
+
+        try:
+            print("..running the NameError...")
+            num = random.random()
+            print(num)
+            print("Apparently no errors")
+            break
+        except NameError:
```
PythonCHB: we weren't expecting you to catch the errors -- just make then, so you got familiar with the core errors you'll get when you code. We'll get into Exception handling an a few weeks. But all done well -- above and beyond!

PR #42, opened 2018-10-16
Exercises for lesson 1 and 2

VioletLuo:
I have commited he every file twice by accident. The files are about exercise 1(session01/break_me.py), print grid exercise (Lesson02_Ex1) and FizzBuzzTest.


students/ZidanLuo/FizzBuzzTest.py
```diff
@@ -0,0 +1,13 @@
+def FizzBuzz():
+	for i in range(1, 101):
```
PythonCHB: Indentation should be four spaces in Python -- not more, and not <tab> characters.

Make sure your editor is set up to put in four spaces when you hit the <tab> key.

students/ZidanLuo/FizzBuzzTest.py
```diff
@@ -0,0 +1,13 @@
...
+		if i % 3 == 0:
```
PythonCHB: can you clean this up to be a little less redundant? 

<hint>

if you do:

`print("something",  end='')`

you won't get a newline after the "something".

students/ZidanLuo/Lesson02_Ex1.py
```diff
@@ -0,0 +1,23 @@
+import sys
```
PythonCHB: this and FizzBuzz are due next week -- more coments then.

students/ZidanLuo/session01/break_me.py
```diff
@@ -0,0 +1,18 @@
+def NameError ():
+    print(x)
+
+# Passing arguments of the wrong type should result in a TypeError
+def TypeError(x):
+    x = x + 1
+
+def SyntaxError():
```
PythonCHB: I doin't see a syntax error here -- I do see a TypeError though...

A syntax error will get caught when you try to run the code, before it even starts running.

students/ZidanLuo/session01/break_me.py
```diff
@@ -0,0 +1,18 @@
...
+    print("error" + 1 + ' ')
+
+def AttributeError(x,y):
```
PythonCHB: this doesn't raise an Attribute Error.

If you call it with one argument, you get a TypeError

VioletLuo:
fixed problems mentioned in the comments

VioletLuo:
Moved all the files into different session folders.

students/ZidanLuo/session02/printGrid.py
```diff
@@ -0,0 +1,23 @@
+import sys
+
+def print_grid(x,y):
+    count = 1
+    #count all the possible number of lines
+    lines = (y + 1) * x + 1
+    plus = "+"
```
PythonCHB: is there any poiint in defining a variable that is a single character anyway?

Maybe -- if you want to be able to change that character later in one place. But in that case, maybe give it a more generic name, like "corner" rather than "plus".

students/ZidanLuo/session02/printGrid.py
```diff
@@ -0,0 +1,23 @@
...
+    minus = "-"
+    shu = "|"
+    empty = " "
+    #draw the horizontal and vertical lines
+    Horizontal = (plus + y * minus) * x +plus
+    Vertical = (shu + y * empty) * x + shu
+    
+    #Use while loop to loop through all the possible lines
+    while count <= lines :
```
PythonCHB: better to use a for loop:

```
for count in range(lines):
    ...
```

students/ZidanLuo/session02/printGrid.py
```diff
@@ -0,0 +1,23 @@
...
+    	if (count % (y + 1)) == 1:
+             print (Horizontal)
+    	else:
+    	     print (Vertical)
+    	count += 1
+#Take arguments in Command Lines
+print_grid(int(sys.argv[1]),int(sys.argv[2]));
```
PythonCHB: nice use of `sys.argv`!

students/ZidanLuo/session02/series.py
```diff
@@ -0,0 +1,40 @@
+import sys
```
PythonCHB: careful with your indentation!!

you want four spaces per indent. always.

Make sure your editor is set up to do that when you hit the <tab> key...

students/ZidanLuo/session02/series.py
```diff
@@ -0,0 +1,40 @@
...
+	if n == 0:
+		return 2
+	elif n == 1:
+		return 1
+	else:
+		return lucas(n-1) + lucas(n-2)
+
+#n if the required parameter to determine which element in the series to print
+#y and z are optional parameters to determine the default values
+def sum_series(n, y = 0, z = 1):
```
PythonCHB: minor style issue, better to not use space around the `=` in parameters:

`def sum_series(n, y=0, z=1):`

See PEP8:

https://www.python.org/dev/peps/pep-0008/

students/ZidanLuo/session02/series.py
```diff
@@ -0,0 +1,40 @@
...
+	if n == 0:
+		return y
+	elif n == 1:
+		return z
+	else: 
+		return sum_series(n-1, y, z) + sum_series(n-2, y, z)
+
```
PythonCHB: Perfect!

students/ZidanLuo/session02/series.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+#print(fibonacci(int(sys.argv[1])))
+#print(lucas(int(sys.argv[1])))
+if len(sys.argv) < 3:
```
PythonCHB: THis would be a good place for a:

```
if __name__ == "__main__":
    put stuff here...
```
block.

PR #43, opened 2018-10-16
break_me_assignment

vanulor:
Here is my break me code

### DIFF
--- a/students/BrianNisonger/break_me.py
+++ b/students/BrianNisonger/break_me.py
@@ -1,0 +1,18 @@
+def NameError():
+	print (x)
+
+def TypeError():
+	a=5;
+	b="17"
+	return a+b
+#def SyntaxError():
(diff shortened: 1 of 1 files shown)


students/BrianNisonger/break_me.py
```diff
@@ -0,0 +1,18 @@
+def NameError():
+	print (x)
+
+def TypeError():
+	a=5;
```
PythonCHB: what the semi-colon doing there? it will be ignored, but no reason to put it there.

students/BrianNisonger/break_me.py
```diff
@@ -0,0 +1,18 @@
...
+	b="17"
+	return a+b
+#def SyntaxError():
+	#for(i;i++;i=0)
```
PythonCHB: funny how you can't write C and expect a Python interpreter to understand it :-)

PR #45, opened 2018-10-16
README

Adolphe76:
this is my first pull request


Adolphe76:
first small assignment

students/Adolphe_N/.project
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
PythonCHB: Did you add this .project file on purpose? there is some debate about whether this kind of IDE file should be in source control...

if you son't want to manage them with git, you can do:

`git rm .project .pydevproject`

PythonCHB:
I'm merging, then I'll put that README file back the way it was in the main repo.

when you next do a "git pull upstream master" you should get it.

students/Adolphe_N/.project
Adolphe76: No it was not on purpose! i think i might have generated them while i was trying to make python 3.7 interpreter works because i had 2.7 already installed. i will remove them. Thanks

Adolphe76: No it was not on purpose! i think i might have generated them while i was trying to make python 3.7 interpreter works because i had 2.7 already installed. i will remove them. Thanks

PR #46, opened 2018-10-16
Adding directory and readme.txt file

dosanjhm:


PythonCHB:
The README in your dir is perfect (though you might put *something* in it :-) )

But you have an extra file: uploadme.txt you've added to the root of the repo. YOu'll need to remove that:

`git rm uploadme.txt`

should do it

PR #49, opened 2018-10-17
break_me.py

Misseswaddles:
My first pass at break_me.py

Best,
Carol

### DIFF
--- a/students/C_Farris/session01/break_me.py
+++ b/students/C_Farris/session01/break_me.py
@@ -1,0 +1,57 @@
+#!/usr/bin/env python3
+
+#Author: Carol Farris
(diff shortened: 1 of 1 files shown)


students/C_Farris/session01/break_me.py
```diff
@@ -0,0 +1,57 @@
...
+        print(val)
+    except NameError:
+        print("Name Error. Did you define your variable?")
+
+
+#the function below will throw a type error if the supplied variable
+#is not valid for the action
+def cause_type_error(val):
+	try: 
+		print(val//2)
```
charlesrobison: Indention should be 4 spaces - try taking a look at the set up in your editor.

PR #54, opened 2018-10-18
Session02 Assignment files

CodeChameleon:
Added the print_grid, fizzbuzz, and series assignment files.


students/DrewSmith/session02/fizzbuzz.py
```diff
@@ -0,0 +1,10 @@
+def fizz_buzz():
```
PythonCHB: very nice!

love how you didn't repeat the tests for %3 and %5.

students/DrewSmith/session02/print_grid.py
```diff
@@ -0,0 +1,34 @@
...
+    print(divide)
+    for __ in range(3):
+        for __ in range(4):
+            print(length)
+        print(divide)
+
+def print_grid(n):
+    square_width = ""
+    for index in range(n):
+        square_width += " " if index % 2 == 0 else "-"
```
PythonCHB: interesting! can you figure a way to do it without the loop -- maybe some multipliation?

also -- does it work right for both even and odd n?

students/DrewSmith/session02/print_grid.py
```diff
@@ -0,0 +1,34 @@
...
+    divide =  "+" + square_width + "+" + square_width + "+"
+    length = "|" + (" " * n) + "|" + (" " * n) + "|"
+    print(divide)
+    for __ in range(3):
+        for __ in range(n):
+            print(length)
+        print(divide)
+
+def print_grid2(grid, size):
+    divide = "+"
```
PythonCHB: I like this -- easy to change the symbols if you want.

students/DrewSmith/session02/series.py
```diff
@@ -0,0 +1,85 @@
...
+"""
+Module for generating adding sequence index values.
+"""
+def fibonacci(n):
+    """
+    Returns the nth value in the fibonacci sequence.
+    
+    :param n: the zero based index value in fibonacci sequence to return
+    """
+    first , second, val = 0, 1, 0
```
PythonCHB: no need to define val here.

students/DrewSmith/session02/series.py
```diff
@@ -0,0 +1,85 @@
...
+        first, second = second, val
+    return val
+
+def lucas(n):
+    """
+    Returns the nth value in the lucas sequence.
+    
+    :param n: the zero based index value in lucas sequence to return
+    """
+    first , second, val = 2, 1, 0
```
PythonCHB: did you notice that this is exactly the same as fibonacci except this line?

inspiration for the next one :-)

students/DrewSmith/session02/series.py
```diff
@@ -0,0 +1,85 @@
...
+    
+    :param n: the zero based index value in fibonacci sequence to return
+    """
+    first , second, val = 0, 1, 0
+    if n <= 0:
+        return first
+    elif n == 1:
+        return second
+
+    for __ in range(2, n + 1):
```
PythonCHB: The goal was do do it recursively, rather than with a loop.

But you ended up doing that with sum_series later -- so all good :-)

students/DrewSmith/session02/series.py
```diff
@@ -0,0 +1,85 @@
...
+
+    assert sum_series(0) == fibonacci(0)
+    assert sum_series(1) == fibonacci(1)
+    assert sum_series(2) == fibonacci(2)
+    assert sum_series(9) == fibonacci(9)
+
+    assert sum_series(0,2,1) == lucas(0)
+    assert sum_series(1,2,1) == lucas(1)
+    assert sum_series(2,2,1) == lucas(2)
+    assert sum_series(9,2,1) == lucas(9)
```
PythonCHB: Awesome assertion tests!

PR #55, opened 2018-10-18
session 2 assignments

ericstreit:
complete

### DIFF
--- a/students/ericstreit/session02/print_grid.py
+++ b/students/ericstreit/session02/print_grid.py
@@ -1,0 +1,30 @@
+
+#print grid lesson
+
+# #Part 2
+def print_grid(gridsize):
+    borderline = "+" + "-" *gridsize + "+" + "-" *gridsize + "+"
+    innerline = "|" + " " *gridsize + "|" + " " *gridsize + "|"
+    print(borderline)
+    for i in range(gridsize):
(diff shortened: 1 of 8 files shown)


students/ericstreit/session02/print_grid.py
```diff
@@ -0,0 +1,30 @@
+
```
PythonCHB: you seem to have the same thing in "session02" and "lesson02" folders.

sorry for the inconsistent terminology -- but you can use "session" or "lesson", but no need for both :-)

students/ericstreit/session02/series.py
```diff
@@ -0,0 +1,38 @@
+
+#fibonacci series exercise
+def fibonacci(n):
+    """This is the Fibonacci function that will return the nth value"""
+    a=0
+    b=1
+    #comment out all but the last of the print statements if you don't want to see the count increase
+    #print(a)
+    #print(b)
+    for i in range(n-2):
```
PythonCHB: the goal was to do it recursively -- but this works too :-)

ericstreit:
oops! I somehow completely missed the 'generalizing' section w/ sum_series
function assignment. I'll get that submitted as soon as I understand what
recursion means :)

Thanks for the feedback!

PR #56, opened 2018-10-18
submitting fizzbuzz.py to session02 

VincentAquila:
Submitting fizzbuzz.py assignment to session02.  Tested and works as described.


VincentAquila:
Adding an improved version of the assignment.  Practicing with Git Hub uploading.

PR #59, opened 2018-10-20
session02 homework

dvlupr:
Adding session01 and session02 folders
Adding four exercises
Updating readme
Adding empty test file

### DIFF
--- a/students/JimJenkins/session01/break_me.py
+++ b/students/JimJenkins/session01/break_me.py
@@ -1,0 +1,30 @@
+#!/usr/bin/env python
+
+'''
+Author: Jim Jenkins (dvlupr)
+Date: 10/13/2018
+
+Each function, when called, should cause an exception to happen
+Each function should result in one of the four most common exceptions you’ll find.
+For review: NameError, TypeError, SyntaxError, AttributeError
+
+'''
+# functions that trigger each error
+def fn_nameerror():
+    print nameerror
+
+def fn_typeerror():
+    print('typeerror')
+
+def fn_syntaxerror():
+    print('syntaxerror
+
+def fn_attributeError():
+    print('attributeerror')
+
+# calling the functions
+fn_nameerror()
+#fn_typeerror()
+#fn_syntaxerror()
+#fn_attributeError()
+
--- a/students/JimJenkins/session02/fizzbuzz.py
+++ b/students/JimJenkins/session02/fizzbuzz.py
@@ -1,26 +1,31 @@
 #!/usr/bin/env python
 
-'''
+"""
 Author: Jim Jenkins (dvlupr)
 Date: 10/18/2018
 
-'''
-import sys
+Write a program that prints the numbers from 1 to 100 inclusive.
+But for multiples of three print “Fizz” instead of the number.
+For the multiples of five print “Buzz” instead of the number.
+For numbers which are multiples of both three and five print “FizzBuzz” instead.
+"""
 
-# request how many times the user wants to run FizzBuzz
-y = int(input('How many numbers do you want to FizzBuzz test?: '))
+# request how many times to run FizzBuzz
+y = 100
+
 
 # set the sentinel value
 x = 0
+
+
 # create a while loop looking for the factors of 3 and 5
 while x != y:
-    # check to ensure the number of sequence runs is reasonable
-    if y > 500:
-        sys.exit('Number is too high for the system, please select a lower number')
     x = x + 1
-    if x % 3 == 0:
-        print (x, 'Fizz')
+    if x % 3 == 0 and x % 5 == 0:
+        print(x, 'FizzBuzz')
+    elif x % 3 == 0:
+        print(x, 'Fizz')
     elif x % 5 == 0:
         print(x, 'Buzz')
     else:
-        print(x, 'FizzBuzz')
+        print(x)
--- a/students/JimJenkins/session02/fizzbuzz.py
+++ b/students/JimJenkins/session02/fizzbuzz.py
@@ -1,0 +1,26 @@
+#!/usr/bin/env python
+
+'''
+Author: Jim Jenkins (dvlupr)
+Date: 10/18/2018
+
+'''
+import sys
+
+# request how many times the user wants to run FizzBuzz
+y = int(input('How many numbers do you want to FizzBuzz test?: '))
+
+# set the sentinel value
+x = 0
+# create a while loop looking for the factors of 3 and 5
+while x != y:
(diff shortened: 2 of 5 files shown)


students/JimJenkins/session01/break_me.py
```diff
@@ -0,0 +1,30 @@
...
+Author: Jim Jenkins (dvlupr)
+Date: 10/13/2018
+
+Each function, when called, should cause an exception to happen
+Each function should result in one of the four most common exceptions you’ll find.
+For review: NameError, TypeError, SyntaxError, AttributeError
+
+'''
+# functions that trigger each error
+def fn_nameerror():
```
charlesrobison: Typically you'll want your function names to be in snake_case.  def name_error(): for example.

students/JimJenkins/session01/break_me.py
```diff
@@ -0,0 +1,30 @@
...
+    print nameerror
```
charlesrobison: Missing the parentheses and quotes on this one - I'm getting a SyntaxError

students/JimJenkins/session01/break_me.py
```diff
@@ -0,0 +1,30 @@
...
+
+def fn_typeerror():
+    print('typeerror')
```
charlesrobison: For this, you'd want to create a function that actually produces the error.  Counterintuitive, I know :)

students/JimJenkins/session01/break_me.py
```diff
@@ -0,0 +1,30 @@
...
+
+def fn_syntaxerror():
+    print('syntaxerror
+
+def fn_attributeError():
+    print('attributeerror')
```
charlesrobison: Same thing on these - camel_case and produce actual errors.  Take a look and reach out if you have questions.

students/JimJenkins/session02/fizzbuzz.py
```diff
@@ -0,0 +1,26 @@
...
+    # check to ensure the number of sequence runs is reasonable
+    if y > 500:
+        sys.exit('Number is too high for the system, please select a lower number')
+    x = x + 1
+    if x % 3 == 0:
+        print (x, 'Fizz')
+    elif x % 5 == 0:
+        print(x, 'Buzz')
+    else:
+        print(x, 'FizzBuzz')
```
charlesrobison: This is pretty close - but not quite.  Getting a lot more FizzBuzz that we should be.  Can you think of a way to simplify the logic a bit?

students/JimJenkins/session02/grid_printer.py
```diff
@@ -0,0 +1,32 @@
...
+            print(('|' + ('  ' * x) + ' ') * a + '|')
+    print(('+' + (' -' * x) + ' ') * a + '+')
+
+
+# format the view
+print('-----' * 7)
+
+
+# run the fundtion based on user input
+fn_grid(x,a)
```
charlesrobison: Works well - nicely done

students/JimJenkins/session02/series.py
```diff
@@ -0,0 +1,39 @@
...
+        z = x + y
+        # add the item to the list
+        num_list.append(z)
+        print(num_list)
+
+# ask the user for how many times they want to run the sequence
+b = int(input('How many times would you like to run the sequence?: '))
+
+# run the sequence per the users instructions
+fn_fibonacci(b)
```
charlesrobison: Looks like this one's not quite finished - still working on lucas, sum_series, and tests?

students/JimJenkins/session01/break_me.py
PythonCHB: NOTE: ifthis code runs, then you are running it in Python2 -- you really do want to figure out how to get Python3 running -- they aren't all that different, but we do use Python3 only features in this class.

students/JimJenkins/session02/series.py
PythonCHB: and we're looking for a function that returns the nth value -- not the whole series ....

and try to do it using recursion.

students/JimJenkins/session02/series.py
```diff
@@ -1,39 +1,58 @@
...
-        z = x + y
-        # add the item to the list
-        num_list.append(z)
-        print(num_list)
 
-# ask the user for how many times they want to run the sequence
-b = int(input('How many times would you like to run the sequence?: '))
+def sum_series(n, x=0, y=1):
+    if n == 0:
+        return n == x
```
PythonCHB: This is going to return “True” or “False”.

students/JimJenkins/session02/fizzbuzz.py
```diff
@@ -0,0 +1,31 @@
...
+# request how many times to run FizzBuzz
+y = 100
+
+
+# set the sentinel value
+x = 0
+
+
+# create a while loop looking for the factors of 3 and 5
+while x != y:
```
PythonCHB: A for loop with range() is a better option here.

students/JimJenkins/session02/fizzbuzz.py
```diff
@@ -0,0 +1,31 @@
...
+    x = x + 1
+    if x % 3 == 0 and x % 5 == 0:
+        print(x, 'FizzBuzz')
+    elif x % 3 == 0:
+        print(x, 'Fizz')
+    elif x % 5 == 0:
+        print(x, 'Buzz')
+    else:
+        print(x)
```
PythonCHB: Logic all good!

students/JimJenkins/session02/series.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def lucas(n):
+    if n == 0:
+        return 2
+    elif n == 1:
+        return 1
+    else:
+        return lucas(n-1)+lucas(n-2)
+
```
PythonCHB: Fib and Lucas are perfect :-)

students/JimJenkins/session02/series.py
```diff
@@ -0,0 +1,58 @@
...
+
+def sum_series(n, x=0, y=1):
+    if n == 0:
+        return n == x
+    elif n == 1:
+        return n == y
+    else:
+        return sum_series(n-1)+sum_series(n-2)
```
PythonCHB: You need to
Pass x and y through, or you will get the Fibonacci series every time ..,.

students/JimJenkins/session02/series.py
dvlupr: updated sum_series  
added assertion

PythonCHB:
Merging this baby!

PR #61, opened 2018-10-20
session01 and session02

habtamua:


students/HABTAMU/session02/printer_Grid_Ex.py
```diff
@@ -0,0 +1,51 @@
...
+# |         |         |
+# |         |         |
+# + - - - - + - - - - +
+
+# Option 1
+
+def print_rectangle(row, col) :
+    for i in range(11) :
+        for j in range(11):
+            if ((i in range (0,15,5) and j%5 == 0)) :
```
PythonCHB: Isn’t the checking of in range() redundant with the % check?

In any case, you are creating a new range object each time through the loop. Better to make one and re-use it. Outside the loop:

Fives = range(0,15,5)

Then in loop:

If I in fives:

(On phone, so capitalization wrong...)

habtamua:
Thank you and will make change as your suggestion later.

Have a great weekend

PR #62, opened 2018-10-20
This is going to try out submit homework

thanhnhan7338:
Hi,

I am trying out to submit homework assignment.
I still not complete with other two assignment yet.

Thanks,

Thanh


students/ThanhNhan/README.md
```diff
@@ -0,0 +1,22 @@
+# Class Repository
```
PythonCHB: did you copy this from the parent Dir?

Anyone, not problem -- you can put anything you want in this file -- it's yours :-)

students/ThanhNhan/lesson02/assignment2/fizzbuzz.py
```diff
@@ -0,0 +1,9 @@
+
+x=1
```
PythonCHB: no need to set x to 1 here -- it will be over-written in the for loop anyway.

students/ThanhNhan/lesson02/assignment2/fizzbuzz.py
```diff
@@ -0,0 +1,9 @@
...
+for x in range (100):
+	if (x % 3):
+		print ("Fizz")
+	elsif (x % 5):
```
PythonCHB: elseif is misspelled -- but sure to run your code!

students/ThanhNhan/lesson02/assignment2/fizzbuzz.py
```diff
@@ -0,0 +1,9 @@
...
+		print ("Buzz")
+	else
+		print (x)
```
PythonCHB: what if it's a multiple of 3 and 5?

students/ThanhNhan/lesson02/assignment2/print_grid2.py
```diff
@@ -0,0 +1,21 @@
+def print_grid2(length, minus): 
+    plus = "+"   
+    minus_num = minus * "-"
```
PythonCHB: careful with variable names here -- I would expect a name like "minus_num" to be a number.

students/ThanhNhan/readme
```diff
@@ -0,0 +1 @@
+test
```
PythonCHB: non need for two readmes...

you can remove one with:

`git rm readme`

PythonCHB:
Looking good!

Good job moving your files, too :-), getting the hang of git a bit now ...

PR #72, opened 2018-10-22
added series (fibonacci) exercise for sesson02

Zach-Cooper:
Kind of struggled with this. Thanks for the template to help keep me on track!

### DIFF
--- a/students/ZachCooper/session02/grid_printer.py
+++ b/students/ZachCooper/session02/grid_printer.py
@@ -1,12 +1,16 @@
-#! usr/bin/env python
+#!usr/bin/env python
 
 #Create function for line one of the grid
 def grid_lineOne():
-	print('+', '-', '-', '-', '-', '+', '-', '-', '-', '-', '+')
+	#print('+', '-', '-', '-', '-', '+', '-', '-', '-', '-', '+')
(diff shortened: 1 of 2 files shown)


students/ZachCooper/session02/grid_printer.py
```diff
@@ -1,12 +1,16 @@
-#! usr/bin/env python
+#!usr/bin/env python
```
PythonCHB: you want in initial slash here:

`#!/usr/bin/env python3`

students/ZachCooper/session02/grid_printer.py
```diff
@@ -1,12 +1,16 @@
...
 
 #Create function for line one of the grid
 def grid_lineOne():
-	print('+', '-', '-', '-', '-', '+', '-', '-', '-', '-', '+')
+	#print('+', '-', '-', '-', '-', '+', '-', '-', '-', '-', '+')
+	#Simplified line one to eliminate spaces
+	print("+ " + "- " * 4 + "+ " + "- " * 4 + "+")
```
PythonCHB: yup! that's better.

PythonCHB:
Merging, so the new stuff will be obvious later :-)

PR #74, opened 2018-10-23
Adding Week 2 exercises

kmonster462:
Adding series, fizzbuzz, and print grid - this is my first solo time using github, I hope I do this right...

### DIFF
--- a/examples/Session02/fizz_buzz.py
+++ b/examples/Session02/fizz_buzz.py
@@ -1,0 +1,14 @@
+"""In a list of 1-100, if num is divisible by three, print 'Fizz', if num is divisible by 5, 
+rint 'Buzz', and if the num can be divided by 3 and 5, print 'FizzBuzz'"""
+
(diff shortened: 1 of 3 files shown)


PythonCHB:
you've got a conflict in the examples dir:

examples/Session02/print_grid.py

you do not want to change those at all -- and if you do, don't commit the change!

you can roll it back with:

```
git checkout c03bb5b2 examples/Session02/print_grid.py
```

PythonCHB:
oops -- you've put all your code int he examples dir.

you want to put them in your own dir.

you should be able to do that with a "git mv" (move)

(in the root dir of the repo)

git mv examples/Session02/fizz_buzz.py students/Kate_Koehn/Session02/

and so on for the three files. (print_grid you want to copy, as you should put back the old one with a git checkout...

PythonCHB:
I'm going to close this -- you can re-open when you have it straightened out.

PR #76, opened 2018-10-23
submit Fibonacci Series Exercise series.py

ShujinCao:


students/JackieCao/Lesson02/series.py
```diff
@@ -1,76 +1,104 @@
 # Fibonacci Series
 
 def fibonacci(n):
-"""
-make sure the input n is an integer and larger or equal to 0
-"""
+    """
+    make sure the input n is an integer and larger or equal to 0
+    """
     if n != int(n):
```
PythonCHB: this really isn't the way to do this.

1) if somethin that cannot be turned into and int gets passed in, you get an exception anyway:

"""
In [9]: fibonacci("fred")
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-9-9853496c18d9> in <module>()
----> 1 fibonacci("fred")

<ipython-input-7-952816d0860d> in fibonacci(n)
      3     make sure the input n is an integer and larger or equal to 0
      4     """
----> 5     if n != int(n):
      6         if not isinstance(n,int):
      7             print("n must be integer")

ValueError: invalid literal for int() with base 10: 'fred'

if it CAN be turned into an int, then you get the error, but it could actually work.

I'd just forget the check -- simply do n -  int(n), and it will work when you want it to, and raise an exeption when it shouldn't work.

or don't even bother with that :-) -- it will probably raise an Exception at some point when it should...

PythonCHB:
I'd still like to see a recursive solution, and one that returns a value rather than printing -- but you've got the logic right here.

Thanks for fixing the indentation !

PR #77, opened 2018-10-23
adding series and grid to session02

VincentAquila:
Still not completely understanding the name = main block, but am slowly learning.  Understanding assertions, these seem like a great tool.  Couldn't get the 3rd part of the series to work (in the series file); I think the problem rests with the way I set up variables in the Fibonacci and Lucas series(?).


students/Vincent_Aquila/session02/fizzbuzz.py
```diff
@@ -0,0 +1,27 @@
+"""
```
PythonCHB: solid solution -- but can you figure how how not to check for %3 and %5 twice?

students/Vincent_Aquila/session02/printgrid.py
```diff
@@ -0,0 +1,51 @@
+"""
+Python210A Vincent Aquila  Fall2018
+using Python 3.7.0
+assignment objective: grid
+"""
+
+#Part 1
+#Function for top, middle, and bottom of grid
+def grid_one():
+	print('+', '-', '-', '-', '-', '+', '-', '-', '-', '-', '+')
```
PythonCHB: if you want to print all this will spaces, probablh beter to build up the string, and make it one print.

things like `"- " * 5`  can be helpful.

-CHB

students/Vincent_Aquila/session02/printgrid.py
```diff
@@ -0,0 +1,51 @@
...
+	for i in range(n):
+		print('|', (' '' ' * n), '|', (' '' ' * n), '|')
+	print('+', ('-'' ' * n), '+', ('-'' ' * n), '+')
+print_grid(15)
+
+
+#Part 3
+"""
+print a grid of any height by any width; change the number of dashes and number
+of spaces in the grid function to alter the size of individual squares
```
PythonCHB: any way to make the squares a different size?

students/Vincent_Aquila/session02/series.py
```diff
@@ -0,0 +1,73 @@
...
+    This function should generalize the fibonacci() and the lucas(),
+    so that this function works for any first two numbers for a sum series.
+    Once generalized that way, sum_series(n, 0, 1) should be equivalent to fibonacci(n).
+    And sum_series(n, 2, 1) should be equivalent to lucas(n).
+    """
+    if n == 0:
+        return n0
+    if n == 1:
+        return n1
+    return sum_series(n-1, n0, n0+n1)
```
PythonCHB: so close!

but this is calling the same sum_series -- so  you want to pass n0 and n1 right on through :-)

PythonCHB:
I'm going to merge, so we'll see the changes clearly in the next PR.

PR #79, opened 2018-10-23
Completed series.py

ktbzzz:


students/JaeKim/session02/series.py
```diff
@@ -0,0 +1,65 @@
...
+def lucas(n):
+    """Returns the nth value of the Lucas number series
+    """
+    if n == 0:
+        return 2
+    if n == 1:
+        return 1
+    return lucas(n - 1) + lucas(n - 2)
+
+def sum_series(n, n0=0, n1=1):
```
PythonCHB: you've missed the point here -- you want sum_series to be able to compute a series with any two numbers to start it off -- fib and lucas are only two possibilities...

look at how similar the two functions are -- how would you generalize that to start off with any two numbers?

PythonCHB:
somehow you've ended up with a conflict in your own file --not sure how.

but after you update the sum_series function, if there is still a conflict, then try to fix it in gitHub by clicking the "resolve conflicts" button.

I'll close this now -- you can re-open when you're ready to resubmit

PR #82, opened 2018-10-23
The Fibonacci Series assignment

framllat:
Put in a good stretch of troubleshooting hours. The sum_series was my downfall. It would work for the first 2 lucas numbers of (0,1) - however, the sum_series function would diverge from the lucas series greater than 1.


students/student_framllat/session02/series.py
```diff
@@ -0,0 +1,69 @@
...
+    """This function calculates the nth value of a Lucas Number
+     starting at the zero index
+    """
+
+    if n == 0:
+        return 2
+    elif n == 1:
+        return 1
+    else:
+        return lucas(n - 2) + lucas(n - 1)
```
PythonCHB: these two are perfect.

students/student_framllat/session02/series.py
```diff
@@ -0,0 +1,69 @@
...
+def sum_series(n, n0=0, n1=1):
+
+    """This function attempts to calculate the nth value of a Series 
+     starting at the zero index
+    """
+
+    if n == 0:
+        return n0
+    elif n == 1:
+        return n1
```
PythonCHB: good start...

students/student_framllat/session02/series.py
```diff
@@ -0,0 +1,69 @@
...
+    else:
+        return sum_series(n - 2) + sum_series(n - 1)
```
PythonCHB: and here we have our problem -- when you call this will n > 1, it sill recursively call itself with smaller numbers until it terminates at the "bottom" -- n = 0 or 1. But when you calll it this way -- it has the samller n, but how does it know what n0 and n1 to use?

students/student_framllat/session03/slicer.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/env python3
+
```
PythonCHB: the idea here is to use slicing as much as possible -- which works with all sequences.

So there should be no need to convert to a list.

students/student_framllat/session03/slicer.py
```diff
@@ -0,0 +1,84 @@
...
+def exch_first_last(n):
+
+    """ This will swap first/last element in sequence
+    """
+
+    a = list(n)
+    a[0], a[-1] = a[-1], a[0]
+    if type(a[0]) is str:
+        return ''.join(a)
+    elif type(a[0]) is int:
```
PythonCHB: why should the type of the elements mater at all?

students/student_framllat/session03/slicer.py
```diff
@@ -0,0 +1,84 @@
...
+        return a
+    else:
+        return None
```
PythonCHB: this should never return None -- always a new sequence....

students/student_framllat/session03/slicer.py
```diff
@@ -0,0 +1,84 @@
...
+
+def rm_every_oth(n):
+
+    """ This will remove every other element in sequence
+    """
+
+    a = list(n)
+    del a[1::2]
```
PythonCHB: instead of deleting -- what if you had returned a new sequence with every other element still there?

students/student_framllat/session03/slicer.py
```diff
@@ -0,0 +1,84 @@
...
+
+def thirds(n):
+    last = n[len(n)//3*2:len(n)]
+    middle = n[len(n)//3:len(n)//3*2]
+    first = n[:len(n)//3]
+    return last+first+middle
+
+my_string = "To whom it may concern it is nominal"
+#my_string = "ssssybpnmlwqtttt"
+my_tuple = (12, 33, 19, 66, 101, 9, 88, 102, 33, 119, 200, 0, 32, 4)
```
PythonCHB: what's key about this is that it is a tuple --not that it's elements are integers!

what if we had a tuple of strings?

str_tuple = ("this", "that", "some", "other", "string", "it", "could", "be", "anything")

all these functions should work.

students/student_framllat/session03/slicer.py
```diff
@@ -0,0 +1,84 @@
...
+out_four_tup = four_rm_four(my_tuple)
+
+out_reverse_str = reverse(my_string)
+out_reverse_tup = reverse(my_tuple)
+
+
+out_thirds_str = thirds(my_string)
+out_thirds_tup = thirds(my_tuple)
+
+print(out_str)
```
PythonCHB: printing to see what you get is handy when writing code -- but better to have assertions -- so you can know that ther correct result was returned.

PythonCHB:
Merging, and giving credit for all but partial credit for slicing lab -- you can re-do it for full credit.

PythonCHB:
I don't see a grid printer, either...

PR #83, opened 2018-10-23
Added two exercises, FizzBuzz.py and PrintGrid.py. Note the PrintGrid.py still has a minor issue

ometemey:
print_grid(2, 3) does not align properly.
+ - - - + - - - +
|    |    |
|    |    |
|    |    |
+ - - - + - - - +
|    |    |
|    |    |
|    |    |
+ - - - + - - - +


ometemey:
Just removed, it.
Should show up in the current merge request.

Thanks
obed

    On Tuesday, October 23, 2018, 4:23:01 PM PDT, Christopher H.Barker, PhD <notifications@github.com> wrote:  
 
 
I don’t want to merge until you roll back the .gitignore....

PythonCHB:
oops -- you deleted the file -- which means if I merge, then it will be deleted in the main repo.

you want to roll it back, by checking out a version from before you changed it:

`git checkout 3a33e76350 .gitignore`

if that doesn't work, try a git pull upstream master, and you should get that commit -- or look at your git log and find a commit from before you changed it.

students/Obed_Metemey/Session02/PrintGrid.py
```diff
@@ -0,0 +1,55 @@
...
+print_grid(4)
+
+
+# Part 3
+# This code draws a similar grid function twith a specified number of rows and columns, and with each cell a given size.
+
+
+def vertical(symbol, spaces):
+    for i in range(symbol):
+        print("| " + " " * spaces, end="")
```
PythonCHB: this is doing spaces one character spaces.

students/Obed_Metemey/Session02/PrintGrid.py
```diff
@@ -0,0 +1,55 @@
...
+    print("|")
+
+
+def horizontal(symbol, spaces):
+    for i in range(symbol):
+        print("+ " + "- " * spaces, end="")
```
PythonCHB: this is doing spaces two character spaces: "- "

this is bigger, than before, so it doesn't line up...

PR #85, opened 2018-10-23
Adding Lesson 2 Homework assignments

dosanjhm:


students/ManmeetDosanjh/Series.py
```diff
@@ -0,0 +1,67 @@
...
+		return Fibonacci(n-2) + Fibonacci(n-1)
+
+
+
+
+
+def Lucas(n):
+	"""return the nth value of the Lucaas Numbers with one argument"""
+	if n == 0:
+		return 2
```
PythonCHB: careful with your indents -- 4 spaces -- ALWAYS!

students/ManmeetDosanjh/grid_printer.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+
+
+
+
+
+# Part 2
+
+#?? I couldn't figure this one out.
```
PythonCHB: non need if you got part 3 :-)

PR #92, opened 2018-10-26
Added list_lab.py for session03

Zach-Cooper:
Couple things that I have gotten snagged on for this assignment.
1)To display all fruits that start with "P", do I need to use enumerate here?
2) I am also stuck on Series 3 for this assignment but feel im close.... How can I create a while loop that loops through each fruit in the list and if user says 'no' the fruit is deleted? Right now I can get the loop to ask for the first fruit using the index but cant have the loop continue onto the next? Any subtle clues to push me in the right direction would be much appreciate. I honestly think im over thinking this.
Thanks


students/ZachCooper/session03/list_lab.py
```diff
@@ -0,0 +1,76 @@
...
+#Add another fruit to the list using "+"
+print(fruit + ['Melons'])
+#Add another fruit to list using insert(), then display
+fruit.insert(0, 'Strawberries')
+print(fruit)
+#Display fruits that begin with "P", using a for loop
+for x in fruit:
+	if x == 'P':
+		print(x)
+"""I think you have to use enumerate??"""
```
PythonCHB: nope -- you are very close. The loop is right, the test for "start with P" is wrong.

Hint, the fruits are strings, loop at the string methods... (or use simple indexing -- how can you get the first character of a string?

students/ZachCooper/session03/list_lab.py
```diff
@@ -0,0 +1,76 @@
...
+#Return the number back and fruit name at that index
+print(fruit[num_question])
+#Add another fruit to the list using "+"
+print(fruit + ['Melons'])
+#Add another fruit to list using insert(), then display
+fruit.insert(0, 'Strawberries')
+print(fruit)
+#Display fruits that begin with "P", using a for loop
+for x in fruit:
+	if x == 'P':
```
PythonCHB: four spaces per indent! ALWAYS! no <tab> characters!

students/ZachCooper/session03/list_lab.py
```diff
@@ -0,0 +1,76 @@
...
+		print(x)
+"""I think you have to use enumerate??"""
+
+#Series 2
+#Display list
+print(fruit)
+#Remove the last fruit from list
+del fruit[5]
```
PythonCHB: what if the list were longer? don't hard-code a 5 in -- there is no need.

students/ZachCooper/session03/list_lab.py
```diff
@@ -0,0 +1,76 @@
...
+#Display original fruit list
+print(fruit)
+#`Ask the user for input displaying a line like “Do you like apples?” for each fruit in the list
+#(making the fruit all lowercase). For each “no”, delete that fruit from the list.
+#For any answer that is not “yes” or “no”, prompt the user to answer with
+#one of those two values (a while loop is good here)
+
+question = ("Do you like {}?".format(fruit[0]))
+"""***how do i get list manipulation to where I can filter the listed indexes into the question
+Can I use NESTED if statements to continue to ask questions***"""
```
PythonCHB: you can nest the while loop inside a for loop.

Follow the logic here -- you want to do somethign for each fruit in the list, so that's a for loop:

```
for fruit in fruits:
    ask_yes_no(question)
    ....
```

students/ZachCooper/session03/list_lab.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+def ask_yes_no(question):
+	"""Ask a yes or no question about the fruit"""
+	response = None
+	while response not in ("yes", "no"):
+		response = input(question.lower())
+	if response == "no":
+		del(fruit[0])
```
PythonCHB: you should probably do the deleting outside this function -- have this function return True or False, and then call it and delete if you get a False.

students/ZachCooper/session03/list_lab.py
```diff
@@ -0,0 +1,76 @@
...
+#(making the fruit all lowercase). For each “no”, delete that fruit from the list.
+#For any answer that is not “yes” or “no”, prompt the user to answer with
+#one of those two values (a while loop is good here)
+
+question = ("Do you like {}?".format(fruit[0]))
+"""***how do i get list manipulation to where I can filter the listed indexes into the question
+Can I use NESTED if statements to continue to ask questions***"""
+
+
+def ask_yes_no(question):
```
PythonCHB: putting the in a function is a great idea.

students/ZachCooper/session03/slicing_lab.py
```diff
@@ -0,0 +1,71 @@
...
+assert delete_every_other_item(a_tuple) == (2, 13, 5, 7)
+
+assert remove_first_last(a_string) == "the Seahawks good this "
+assert remvove_first_last(a_tuple) == ()
+
+assert reverse(a_string) == "raey siht doog skwahaeS eht erA"
+assert revers(a_tuple) == (7, 32, 5, 12, 13, 54, 2)
+
+assert slice_seq_thirds(a_string) == d this year Are the Se ahawks goo
+assert slice_seq_thirds(a_tuple) == (5, 32, 7) (2, 54) (13, 12)
```
PythonCHB: This looks like a syntax error — did it run??

PythonCHB:
Cox use some tweaking to the slicing lab — but looking good!

PythonCHB:
you got most of it -- merging -- but would be good to get the list lab figure out :-)

PR #109, opened 2018-10-30
Updated slicing_lab

dtingg:
I updated slicing_lab in the Session03 folder.  I used slicing in all of the exercises.


PythonCHB:
This all looks good.

PR #118, opened 2018-11-04
Mailroom pt 2 and Trigram Maker

rushinstuffin:
Really fun stuff, especially the trigram project. Any feedback would be amazing. Thanks!


students/KyleBarry/session04/trigrammify.py
```diff
@@ -0,0 +1,60 @@
...
+import requests
+from bs4 import BeautifulSoup
+
+
+def get_text(url): 
+    """ 
+    Get text data from url of gutenberg project so as to not have it within code
+    """
+    page = requests.get(url)
+    soup = BeautifulSoup(page.content, "lxml")
```
PythonCHB: getting fancy!

butnote that project gutenbuerg also has plain tet download:

http://www.gutenberg.org/files/2701/2701-0.txt

So you could have done that a bit simpler :-)

(check out the "requests" module for quick and easy downloading -- `pip install requests`)

Also -- Charles had problems with BeautifulSoup -- so couldn't run your code -- not your problem , but it is a reminder why it's good to use as simple as possible tools :-)

students/KyleBarry/session04/trigrammify.py
```diff
@@ -0,0 +1,60 @@
...
+    text = soup.find_all("p")
+    texter = [i.text for i in text]
+    words = "".join(texter).split()
```
PythonCHB: next (optional) step -- do something "smart" with punctuation :-)

students/KyleBarry/session04/trigrammify.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    Select tris dictionary key to start sequence if first letter is
+    capitalized.
+    Iterate for approximate length of text and create tuples of last two words.
+    Add values of tuples that are in tris dictionary, otherwise append random
+    word.
+    """
+    while True:
+        new_text = list(random.choice(list(tris.keys())))
+        if new_text[0][0].isupper():
```
PythonCHB: Ah! looking for an actual start of a sentence? clever!

not a very efficient way to do it though. Maybe have a separate dict (or list) of just the "start of sentence" work pairs?

students/KyleBarry/session04/trigrammify.py
```diff
@@ -0,0 +1,60 @@
...
+    """ 
+    Get text data from url of gutenberg project so as to not have it within code
+    """
+    page = requests.get(url)
+    soup = BeautifulSoup(page.content, "lxml")
+    text = soup.find_all("p")
+    texter = [i.text for i in text]
+    words = "".join(texter).split()
+    # Pass parsed words of book in list form to trigram dictionary maker
+    make_trigrams(words)
```
PythonCHB: better to return the words here, and pass that on outside this function.

Also -- maybe have this function cache the book, or the words, on disk, so you don't have to re-download every time!

students/KyleBarry/session04/trigrammify.py
```diff
@@ -0,0 +1,60 @@
...
+            break
+        else:
+            continue
+
+    for i in range(len(words)-2):
+        last_two = tuple(new_text[-2:])
+        try:
+            new_text.append(random.choice(tris[last_two]))
+        except KeyError:
+            new_text.append(random.choice(new_text))
```
PythonCHB: that's a clever way to get a random word -- though maybe not ideal if you get a key error early on!

students/KyleBarry/session04/trigrammify.py
```diff
@@ -0,0 +1,60 @@
...
+    word.
+    """
+    while True:
+        new_text = list(random.choice(list(tris.keys())))
+        if new_text[0][0].isupper():
+            break
+        else:
+            continue
+
+    for i in range(len(words)-2):
```
PythonCHB: why -2 ? this could really be any length -- no need to be the same as the input text.

students/KyleBarry/session04/trigrammify.py
```diff
@@ -0,0 +1,60 @@
...
+            new_text.append(random.choice(tris[last_two]))
+        except KeyError:
+            new_text.append(random.choice(new_text))
+
+    print(" ".join(new_text))
+
+if __name__ == "__main__":
+    """Passing in Moby Dick for now, but any URL from gutenberg project
+    works"""
+    get_text("http://www.gutenberg.org/files/2701/2701-h/2701-h.htm")
```
PythonCHB: nicely done.

Next step maybe to do more with sentences -- you make sure to kick it off with a capitalized word, but then what?

But full credit -- only do more for fun :-)

students/KyleBarry/session03/mailroom.py
```diff
@@ -1,59 +1,85 @@
...
+        no_dons = [len(v) for v in donors.values()]
+        avg_don = [round(sum(v)/len(v), 2) for v in donors.values()]
+        total_dons = [sum(v) for v in donors.values()]
+        zipped = zip(names, no_dons, avg_don, total_dons)
+
+        for i in zipped:
+            writer.writerow(i)
+
+
+# Do not name this function quit, it leads to stack overflow...
```
PythonCHB: Ahh -- that's overflows because you are calling quit() from within the function.

you will have overridden the built in quit().

Which, by the way I had no idea existed! It didn't used to exist back in the day....

PR #122, opened 2018-11-05
Added Session04 exercises

dtingg:
I added dict_lab, mailroom_two, and trigrams to the Session04 folder.


students/DiannaTingg/Session04/trigrams.py
```diff
@@ -0,0 +1,102 @@
...
+
+# Use dictionary to build new text
+def use_trigrams(d):
+    start = get_random_word(d)
+
+    alternate_word = ""
+
+    text_words = [start[0], start[1]]
+
+    while len(text_words) < 250:
```
charlesrobison: Getting hung up on this function when running this.  Could be the text file I'm using.  This while loop appears to be the culprit.  Let's discuss.

PR #128, opened 2018-11-06
Finished assigments

vanulor:
Finish assignments for this week.  I think there might be a issue with the trigram, it doesn't spit out as many as I think it should, but it does function.  Mailroom is current to this week.


students/BrianNisonger/session3/mailroom.py
```diff
@@ -0,0 +1,51 @@
+#usr/bin/env python3
+
+"""
+mailroom assignment"
+"""
+
+donor_names=("Fred Jones","Amy Shumer")
+donor_donations=([100,200,300],[400,400,400])
```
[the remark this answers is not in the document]
PythonCHB: I"m going to move on the your next version for the rest of the review.

students/BrianNisonger/session3/slicing.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+    a_new_sequence=seq[::-1]
+    return a_new_sequence
+
+
+def reorder_thirds(seq):
+    """
+    Rearranges by thirds (Last,First,Middle) and reverses order of sequences
+    """
+    a_new_sequence=reverse_sequence(seq[-len(seq)//3:])+reverse_sequence(seq[:len(seq)//3])+reverse_sequence(seq[len(seq)//3:-len(seq)//3])
```
PythonCHB: a bit better to save len(seq)//3 and then re-use it, but otherwise perfect.

students/BrianNisonger/session4/dict_lab.py
```diff
@@ -0,0 +1,54 @@
+def dictionary_1(food_dict):
+    print(food_dict)
+    del food_preferences["cake"]
+    print(food_preferences)
+    food_preferences["fruit"]="Mango"
+    print(food_preferences)
+    for items in food_preferences:
+        print (items)
+    for items in food_preferences:
```
PythonCHB: you can do:

`for values in food_preferences.values()`

students/BrianNisonger/session4/dict_lab.py
```diff
@@ -0,0 +1,54 @@
...
+        print (items)
+    for items in food_preferences:    
+        print (food_preferences[items])
+    print(bool("cake" in food_preferences))
+    print(bool("Mango" in food_preferences.values()))
+
+def dictionary_2(food_preferences):
+      count_t={}
+      for items in food_preferences:
```
PythonCHB: careful about names: "items" makes think it's more than one item, when it's only one item in this case.

students/BrianNisonger/session4/file_excercise.py
```diff
@@ -0,0 +1,44 @@
+import pathlib
```
PythonCHB: nice use of pathlib!

students/BrianNisonger/session4/file_excercise.py
```diff
@@ -0,0 +1,44 @@
...
+        outfile.write(infile.read())
+
+def file_3(source):
+    name_dict={}
+    lang_list=[]
+    with open(source) as f:
+        lines=f.readlines()
+    for line in lines:
+        line_out=line.split(":")
+        if line_out[0] !="Name":
```
PythonCHB: This is always going to be at the top of the file, so probably better to skip it somehow first, then you don't have to keep checking.

students/BrianNisonger/session4/file_excercise.py
```diff
@@ -0,0 +1,44 @@
...
+            name_dict[line_out[0]]=line_out[1]
+    for name in name_dict:
+        if name_dict[name] !="":
```
PythonCHB: name_dict[name] is the value. If you want to loop through the keys and values of a dict, then:

```
for name, values in name_dict.items():
    ...
```

will do it.

students/BrianNisonger/session4/file_excercise.py
```diff
@@ -0,0 +1,44 @@
...
+            languages=(name_dict[name].split(","))
+            for language in languages:
+                if language.islower():
+                    language=language.strip()
+                    lang_list.append(language)
+    c=collections.Counter(lang_list)
```
PythonCHB: nice find!

though rather than building up the list, and then passing it to counter, you could be adding stuff to the counter as you loop through the file.

students/BrianNisonger/session4/mailroom.py
```diff
@@ -0,0 +1,91 @@
+import statistics
```
PythonCHB: Nice find!

students/BrianNisonger/session4/mailroom.py
```diff
@@ -0,0 +1,91 @@
+import statistics 
+
+def sort_key(item):
+    return item[1]
+    
+def thank_you():
```
PythonCHB: all good logic, but too much activity all in one function -- you want to break it up so that you can develop and test the pieces separately.

This will become more clear when you try to test the code :-)

students/BrianNisonger/session4/mailroom.py
```diff
@@ -0,0 +1,91 @@
...
+            donation=float(donation)
+            donors[full_name].append(donation)
+            thank_you_string=f'Dear {full_name}, Thank you for your donation of {donation:.2f}.  These funds help save the migratory butterflies of New South East.  Thank you'
+            file_name=full_name.replace(" ","_")
+            file_name=f'{file_name}.txt'
+            with open(file_name,"w") as outfile:
+                outfile.write(thank_you_string)
+
+def report():
+    sum_donor={}
```
PythonCHB: rather than a separate dict ofr each, better to have one dict, and a structure(tuple, another dict) that gets stored for each donor.

students/BrianNisonger/session4/mailroom.py
```diff
@@ -0,0 +1,91 @@
...
+    sum_donors=[]
+    headers=("Donor Names","Total Given","Num Gifts","Average Gifts")
+    header_string=f'{headers[0]:<15} {headers[1]:<15} {headers[2]:<15} {headers[3]:<15}'
+    print(header_string)
+    for donor in donors:
+        sum_donor[donor]=sum(donors[donor])
+        num_donations[donor]=len(donors[donor])
+        average_donor[donor]=statistics.mean(donors[donor])
+    for donor_name in sum_donor:
+        sum_donors_tuple=[donor_name,sum_donor[donor_name]]
```
PythonCHB: Why not create this tuple in the first place?

students/BrianNisonger/session4/trigram.py
```diff
@@ -0,0 +1,59 @@
...
+"""
+def make_words(source):
+    words=""
+    with open(source) as f:
+        lines=f.readlines()
+    punct=[",",".","-","(",")","\n","!","'s","'","?","_",";"]   
+    for line in lines:
+        for punc in punct:
+            line=line.replace(punc," ")
+        words=words+line
```
PythonCHB: do you get a space between the lines?

But anyway -- it is inefficient to build up a long string by adding pieces one at a time. Better to process each line into a list of words, and accumulate that in a list -- you want a list in the end anyway.

-CHB

students/BrianNisonger/session4/trigram.py
```diff
@@ -0,0 +1,59 @@
...
+    else:
+        print(story)
+        
+    
+            
+    
+source="ghosts.txt"
+words=make_words(source)
+tris=make_trigrams(words)
+new_words=input("Please enter new_words:")
```
PythonCHB: so how would you get a starting point form the trigrams dict?

students/BrianNisonger/session4/trigram.py
```diff
@@ -0,0 +1,59 @@
...
+            tris[pair].append(w3)
+        else:    
+            tris[(w1, w2)]=[w3]
+    return tris
+
+def make_story(tris,new_words,story):
+    new_list=new_words.split()
+    new_tuple=tuple(new_list)
+    if new_tuple in tris:
+        next_word=str(tris[new_tuple])
```
PythonCHB: tris[new_tuple] should be a list of words -- you only want one of them.

students/BrianNisonger/session4/trigram.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def make_trigrams(words):
+    tris ={}
+    for w1,w2,w3 in zip(words[:-2],words[1:-1],words[2:]):
+        pair=(w1,w2)
+        if pair in tris:
+            tris[pair].append(w3)
+        else:    
+            tris[(w1, w2)]=[w3]
```
PythonCHB: `trids[pair]` ?

students/BrianNisonger/session4/trigram.py
```diff
@@ -0,0 +1,59 @@
...
+    return tris
+
+def make_story(tris,new_words,story):
+    new_list=new_words.split()
+    new_tuple=tuple(new_list)
+    if new_tuple in tris:
+        next_word=str(tris[new_tuple])
+        punct=[",",".","-","(",")","\n","[","]","'"]   
+        for punc in punct:
+            next_word=next_word.replace(punc,"")
```
PythonCHB: Punctuation should already be stripped out, yes?

students/BrianNisonger/session4/trigram.py
```diff
@@ -0,0 +1,59 @@
...
+    tris ={}
+    for w1,w2,w3 in zip(words[:-2],words[1:-1],words[2:]):
+        pair=(w1,w2)
+        if pair in tris:
+            tris[pair].append(w3)
+        else:    
+            tris[(w1, w2)]=[w3]
+    return tris
+
+def make_story(tris,new_words,story):
```
PythonCHB: you want the next word to be chosen based on the last 2 words in the story.

so you should only need to pass in the story here -- you can kick it off with a two word "story"

students/BrianNisonger/session4/trigram.py
```diff
@@ -0,0 +1,59 @@
...
+    new_list=new_words.split()
+    new_tuple=tuple(new_list)
+    if new_tuple in tris:
+        next_word=str(tris[new_tuple])
+        punct=[",",".","-","(",")","\n","[","]","'"]   
+        for punc in punct:
+            next_word=next_word.replace(punc,"")
+        new_words=f'{new_tuple[1]} {next_word}'
+        story=f'{story} {next_word}'
```
PythonCHB: the story is always going to be two words???

students/BrianNisonger/session4/trigram.py
```diff
@@ -0,0 +1,59 @@
...
+        tris1=tris
+        make_story(tris1,new_words,story)
+    else:
+        print(story)
+
```
PythonCHB: The logic is almost, but not quite what we were going for here.

I'll give you partial credit.

PR #129, opened 2018-11-06
session 4 homework

dvlupr:


dvlupr:
Changes to mailroom to get this going
Adding file_exercise to main branch

students/JimJenkins/session04/file_exercise.py
```diff
@@ -0,0 +1,59 @@
...
+cwd = os.getcwd()
+files = os.listdir(cwd)
+
+for file in files:
+    print(cwd + '\\' + file)
+
+
+#copy file
+ori_file = 'students.txt'
+src_file = open(ori_file, 'r')
```
PythonCHB: open opens in text mode be default -- so this will nly work on text files -- it may barf if you try to use this with a  binary file.

try a  JPEG or something as an example.

students/JimJenkins/session04/file_exercise.py
```diff
@@ -0,0 +1,59 @@
...
+    words = line.split()
+    for word in words:
+        counts[word] = counts.get(word, 0) + 1
+
+for key, val in counts.items():
+    if key == 'Swift':
+        counts.pop('Swift')
+    print(key, val)
+
+#todo need to determine how to remove the names and nicknames
```
PythonCHB: I was about to say that :-)

PR #131, opened 2018-11-06
Lots of hours and research

framllat:
Still working hard understand the concepts. Thanks for the patience.


students/student_framllat/session04/mailroom2.py
```diff
@@ -0,0 +1,247 @@
+#!/usr/bin/env python
+"""
+mailroom assignment
+
+This version uses a dict for the main db, and exception handling to
+check input
+
+*** NOTE *** once again significant utilization of provided solution
+utilized to create functioning program. Added some modifications in
+order to grasp the concepts.
```
PythonCHB: that's a good way to do it!

students/student_framllat/session04/mailroom2.py
```diff
@@ -0,0 +1,247 @@
...
+def find_donor(name):
+    """
+    Find a donor in the donor db
+
+    :param: the name of the donor
+    :returns: The donor data structure -- key is lowercase name with hyphens
+    instead of spaces - ex: Lex Luther = lex-luther
+    None if not in the donor_db
+    """
+    key = name.replace(" ", "-").strip().lower()
```
PythonCHB: since you need this in two places -- and the really shold be the same -- make a function that does it:

```
def name_to_key(name):
    name.replace(" ", "-").strip().lower()
```

Did I do that in my solution? I should have :-)

students/student_framllat/session04/mailroom2.py
```diff
@@ -0,0 +1,247 @@
...
+                      "2": print_donor_report,
+                      "3": save_letters_to_disk,
+                      "4": quit}
+
+    while True:
+        selection = main_menu_selection()
+        try:
+            selection_dict[selection]()
+        except KeyError:
+            print("\nError: menu selection is invalid!")
```
PythonCHB: all looking good -- but do make sure you "get" it all -- particularly the dispatch dict for the main menu!

students/student_framllat/session04/trigrams.py
```diff
@@ -0,0 +1,38 @@
...
+"""
+trigrams
+"""
+
+# import file into a list
+with open('./four_sc.txt') as f:
+    #word_list = ()
+    word_list = []
+    lines = f.readlines()
+    for l in lines:
```
PythonCHB: you can do:
```
    for l in f:
```

and it will do the same thing, without loading it into memory all at once.

students/student_framllat/session04/trigrams.py
```diff
@@ -0,0 +1,38 @@
...
+            tris[pair].append(w3)
+        else:
+            tris[pair] = [w3]
+
+    return tris
+
+print(word_list)
+#print(len(word_list))
+#t_word_list = tuple(word_list)
+tris = make_trigrams(word_list)
```
PythonCHB: now to make new text!

framllat:
Hi Chris --


I put in the suggested name_to_key function. But first I did my best to use pytest...


However, my latest stab for test_mailroom2.py:


import mailroom2 as mr2
#from mailroom2 import find_donor


def test_find_donor():
    donor_db = mr2.get_donor_db()
    t_name = "Oscar Madison"
    result = mr2.find_donor(t_name)
    assert result == ('Oscar Madison', [900, 450, 1200])


Fails miserably -- basically I continue to get stumped on the donor_db - I tried in the beginning not even using it.

However, I keep receiving a NameError: referencing the donor_db..


> return donor_db.get(key)
E NameError: name 'donor_db' is not defined

mailroom2.py:68: NameError


I'm thinking there is an additional import to get access to the mailroom2.py donor_db ?


Thanks for any insight.

PythonCHB:
nope, not an import.

The functions in  mailroom2.py expect donor_db to be in the global (that is global to that file) namespace. 

The file works when run because of this line code:
```
if __name__ == "__main__":
    donor_db = get_donor_db()
```
That is creating donor_db in the global namespace.

But when you import the module from the test code,  __name__ is not  "__main__", so that code doesn't run, and donor_db is not defined.

When you create donor_db in your test function, it is a name in that function, but the functions in the mailroom2 file do not have access to it.

The easiest way to fis this is to move the line:

`donor_db = get_donor_db()`

outside of the if __name__ == "__main__": block, and then it will always run.

you could also put it in teh module namespace in our test code:

```
import mailroom2 as mr2

mr2.donor_db = mr2.get_donor_db()
```
or even inject a different donor_db for testing.

HTH,

-CHB

PR #136, opened 2018-11-11
Mailroom Part 2 is ready to be graded

timothyjpauley:
This file is ready to be graded. However, there is a bug in the Thank-you letter menu when you have a donor print a list. I am working on that.


students/TimPauley/session04/MailRoomPart2.py
```diff
@@ -0,0 +1,146 @@
...
+    
+def main_menu_selection():
+    action = input(dedent('''
+        Choose an action:
+        1 - Send a Thank You
+        2 - Create a Report
+        3 - Send letters to everyone
+        4 - Quit
+        > '''))
+    return action.strip()
```
PythonCHB: you should't need to strip a string you just made ...

students/TimPauley/session04/MailRoomPart2.py
```diff
@@ -0,0 +1,146 @@
...
+    print(gen_letter(donor))
+
+
+def sort_key(item):
+    return item[1]
+
+
+def generate_donor_report():
+    report_rows = []
+    for (name, gifts) in donor_db.values():
```
PythonCHB: nice!

PythonCHB:
I don't see a problem with the Thank-you letter menu -- did you fix that?

PR #138, opened 2018-11-12
Submitting Session 5 Homework

kmonster462:
Session 5 Homework - didn't add any list comprehensions because the only for loops I have are all print statements, so I wasn't sure how to use list comprehension with them. Also adding dict_lab from Session 4.


PythonCHB:
Good call on the list comps -- they are NOT the right solution to loops with prints :-)

PR #140, opened 2018-11-12
Added Session05 exercises and fixed trigrams in Session04

dtingg:
I added except_exercise, except_test, and mailroom_three to the Session05 folder.  
I also fixed a bug in trigrams for Session04.


students/DiannaTingg/Session04/trigrams.py
```diff
@@ -23,30 +20,26 @@ def make_trigrams(words):
...
         else:
-            while pair not in d:
-                alternate_word = get_random_word(d)
+            alternate_word = get_random_pair(d)
 
-            text_words.append([alternate_word[0], [alternate_word[1]]])
+            text_words.append(list(alternate_word))
 
     text_words[0] = text_words[0].capitalize()
     final_text = " ".join(text_words) + "!"
```
PythonCHB: nice improvements!

students/DiannaTingg/Session05/except_exercise.py
```diff
@@ -0,0 +1,54 @@
...
+langs = ['java', 'c', 'python']
+
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    more_joke = more_fun(langs[1])
+else:
+    more_fun(langs[2])
+finally:
+    last_fun()
```
PythonCHB: perfect!

students/DiannaTingg/Session05/mailroom_three.py
```diff
@@ -0,0 +1,181 @@
...
+        # If user enters an empty string
+        except IndexError:
+            pass
+
+        # Get menu selection from dictionary
+        try:
+            menu_dict.get(answer)()
+
+        # If user types an invalid letter or number
+        except TypeError:
```
PythonCHB: what would raise a TypeError here?

PythonCHB: ah - got it -- if get() defaults to None -- so you'd get a TypeError if it's not in the dict.

That works!

PR #154, opened 2018-11-14
HW5 Complete Mailroom and Exception

dancornutt:
I can clean up mailroom a bit. I'd like to get the report format string a little cleaner with comprehension but I couldn't figure it out in time this week. Also still iffy on where and when I should use exceptions.


students/DanCornutt/session05/dict_set_lab2.py
```diff
@@ -0,0 +1,52 @@
...
+        food_prefs['fruit'], food_prefs['salad'],
+        food_prefs['pasta'])
+    return body
+
+
+def hex_dict():
+    """creates dictionary of integers and equivalent hexidecimal numbers
+
+    """
+    h_dict = dict(list(enumerate([hex(i) for i in range(15)])))
```
PythonCHB: no need for the list() here -- enumerate already is an iterable.

And can you do it with a dict comp?

students/DanCornutt/session05/dict_set_lab2.py
```diff
@@ -0,0 +1,52 @@
...
+    return h_dict.items()
+
+
+def a_dict():
+    """creates dictionary of number of A's from c_dict.values()
+    """
+    a_dict_count = dict(c_dict())
```
PythonCHB: calling c_dict already returns a new dict -- not need to wrap dict()around it.

students/DanCornutt/session05/dict_set_lab2.py
```diff
@@ -0,0 +1,52 @@
...
+    for k, v in a_dict_count.items():
```
PythonCHB: can you do this with a dict comp?

students/DanCornutt/session05/except_exercise.py
```diff
@@ -0,0 +1,55 @@
...
+langs = ['java', 'c', 'python']
+
+try:
+    more_fun(langs[0])
+except IndexError:
+    more_fun(langs[1])
+else:
+    more_fun(langs[-1])
+finally:
+    last_fun()
```
PythonCHB: nice job!

students/DanCornutt/session05/exceptions_lab.py
```diff
@@ -0,0 +1,17 @@
...
+        : returns : input from user or None if error
+    """
+    try:
+        return input()
+    except (EOFError, KeyboardInterrupt):
+        return None
+
+
+if __name__ == "__main__":
+    safe_input()
```
PythonCHB: well done.

students/DanCornutt/session05/mailroom.py
```diff
@@ -0,0 +1,209 @@
...
+
+def write_donor(name):
+    """Writes thank you email to file"""
+    text = """Dearest {donor}, \n
+    We greatly thank you for your recent contribution of ${recent:.2f}.\n
+    It will go straight to the person who needs it the most, our CEO.\n
+    Please give more next time.\n
+    \tLove,\n
+    \t\t\tThe Team""".format(donor=name, recent=DONORS[name][-1])
+    with open(name.replace(" ","") + "_thank_you.txt", 'w') as f_out:
```
PythonCHB: better to genreate the letter and write to disk in separate functions.

students/DanCornutt/session05/mailroom.py
```diff
@@ -0,0 +1,209 @@
...
+        donor = input("Type Donor Name =>").title()
+        if check_donor(donor):
+            write_donor(donor)
+        else:
+            print("Sorry I could not find the donor, exiting...")
+
+
+def donor_db():
+    """Editing Donor Menu"""
+    menu_selection(DATABASE_PROMPT, DATABASE_DISPATCH)
```
PythonCHB: better to define these in this function, rather than as globals.

students/DanCornutt/session05/mailroom.py
```diff
@@ -0,0 +1,209 @@
...
+            len_col["Tt_size"] = len(str(sum(d[1])))
+        if len((d[1])) > len_col["ng_size"]:
+            len_col["ng_size"] = len((d[1]))
+        if len(str(sum(d[1])/len(d[1]))) > len_col["ag_size"]:
+            len_col["ag_size"] = len(str(sum(d[1])/len(d[1])))
+        rpt_sheet.append((d[0], sum(d[1]), len(d[1]), sum(d[1])/len(d[1])))
+    rpt_sheet.sort(key=return_total, reverse=True)
+
+    sheet = (
+        "{nm:{mnm}} | {tot:<{mtot}} | {ng:<{mng}} | {ag:<{mag}}\n{header}".format(
```
PythonCHB: kind of a lot in one "line" of code -- maybe put all that data in a dict -- then pass it in to format with **

students/DanCornutt/session06/mailroom.py
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+DATABASE_DISPATCH = {
+    "1": donor_list,
+    "2": edit_donor,
+    "3": quit_menu
+}
+
+if __name__ == "__main__":
+    menu_selection(WELCOME_PROMPT, MAIN_DISPATCH)
```
PythonCHB: this looks pretty much the same -- I assume a copy for testing next lesson.

But note: you still need to clean up the print() and input() a bit more -- they should be isolated in functions that do almost nothing else, so you can test as much as possible.

students/DanCornutt/session06/test_cigar_party.py
```diff
@@ -0,0 +1,61 @@
...
+def test_9():
+    assert cigar_party(39, False) is False
+
+
+def test_10():
+    assert cigar_party(40, True) is True
+
+
+def test_11():
+    assert cigar_party(39, True) is False
```
PythonCHB: nice job on these!

students/DanCornutt/session06/test_mailroom.py
```diff
@@ -0,0 +1,127 @@
...
+    """Finds donor in list"""
+    assert check_donor("Billy Bills") is True
+
+def test_check_donor_2():
+    """Does not find donor in list"""
+    assert check_donor("Billy Billz") is False
+
+def test_gen_letter_1():
+    """Generates valid letter for donor in list"""
+    assert gen_letter("Billy Bills") == """Dearest Billy Bills,
```
PythonCHB: this works, but will fail if the template changes at all.

can you test some of the contents of the letter, without checking the entire text?

students/DanCornutt/session06/test_mailroom.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Please give more next time.
+
+    	Love,
+
+    		The Team"""
+
+def test_gen_letter_1():
+    """Does not generate valid letter for donor not in list"""
+    with pytest.raises(KeyError):
```
PythonCHB: note that this is going to pass or fail based on a KeyError or not being raised -- so you don't need to actually check anything about the result.

students/DanCornutt/session06/test_mailroom.py
```diff
@@ -0,0 +1,127 @@
...
+    It will go straight to the person who needs it the most, our CEO.
+
+    Please give more next time.
+
+    	Love,
+
+    		The Team"""
+
+def test_write_donor(): #Tests if file has been created in directory
+    write_donor("Billy Bills")
```
PythonCHB: nice!

students/DanCornutt/session06/test_mailroom.py
```diff
@@ -0,0 +1,127 @@
...
+
+def test_add_donation():
+    #TODO Should be refactored to return string vs print to screen.
+    pass
+
+def test_add_money():
+    #not sure how to test for user input in pytest.
+    pass
+
+def test_thank_you_all():
```
PythonCHB: good test, but pretty sensitive to the test dataset.

PythonCHB:
you have a conflict:

Conflicting files
students/DanCornutt/session05/mailroom.py

try doing a:

git pull upstream master

and resolve them in your repo, the push, and see if that cleans it up.

charlesrobison:
This is ready to merge - still need to review recent commits for new assignments.

dancornutt:
Any action needed on my side of things? I used the browser to resolve conflicts.

students/DanCornutt/session08/circle.py
```diff
@@ -0,0 +1,108 @@
...
+
+    @classmethod
+    def from_diameter(cls, diameter):
+        """Alternate constructer for from_diameter
+        param1 : diameter of wanted circle"""
+        return cls(diameter / 2)
+
+
+    def __str__(self):
+        return "{} with radius {:.6f}".format(self.name, self.radius)
```
PythonCHB: ahh! got it -- using name attribute sothat subclasses will work with the same __str__.

you can get that with metaprogramming trickery:

```
In [61]: class C: 
    ...:     def test_name(self): 
    ...:         print("class name is:") 
    ...:         print(self.__class__.__name__) 
    ...:                                                                        

In [62]: c = C()                                                                

In [63]: c.test_name()                                                          
class name is:
C
```

we were not expecting you to figure that out at this point :-)

students/DanCornutt/session08/circle.py
```diff
@@ -0,0 +1,108 @@
...
+    def __add__(self, other):
+        try:
+            new = Circle(self.radius + other.radius)
+        except AttributeError:
+            new = Circle(self.radius + other)
+        return str(new)
+
+
+    def __radd__(self, other):
+        new = Circle(self.radius + other)
```
PythonCHB: not that freidnly to subclassing -- if you add two Spheres, you'll get a Circle :-)

you can use `self.__class__` instead.

Again, didn't really expect you to figure that out.

PythonCHB:
lost track a bit of what we reviewed.

so merging this, and if you have any questions about something we didn't comment on --send them along.

PR #155, opened 2018-11-14
Except exercise

Misseswaddles:
Pushing except exercise from new branch with no git issues, hurray!


students/C_Farris/session04/Mailroom_2.py
```diff
@@ -188,14 +184,4 @@ def retry():
...
-
-
-
-
-
-
-
-
-
+    main()
```
PythonCHB: this is a bit odd that it's in this branch -- but no real changes, it may be from our attempt to clean things up.

not a biggie.

students/C_Farris/session05/except_exercise.py
```diff
@@ -0,0 +1,59 @@
...
+langs = ['java', 'c', 'python']
+
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    more_joke = more_fun(langs[1])
+finally:
+    lastItem = len(langs) - 1
+    more_fun(langs[lastItem])
+    last_fun()
```
PythonCHB: perfect !

PythonCHB:
after I merge, you can do a:

git pull upstream master 

and you'll get this in your master.

make sure you are on your master branch when you do that!

PR #156, opened 2018-11-15
Completed Mailroom 3

Misseswaddles:


students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+
+def send_file_to_disk(person, myfile):
+    """
+    Accepts test file and writes to disk
+    :param (myfile):
+    :return: none. Will send file to directory set in outpath.
+    """
+    filename = person.replace(' ', '_') + '.txt'
+    filename = os.path.join(OUT_PATH, filename)
+    open(filename, 'w').write(myfile)
```
PythonCHB: this is writing a string, which you want to write to a file, not an actual file.

nothing wring with the code, but the docs should be clear on this.

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+                       getDonation='No cash specified'):
+    """
+    assembles a thank you letter
+    :param(getPerson):
+    :param(getDonation):
+    :return: assembled text file with persons full name.txt
+    """
+    return dedent(
+        '''\tDear {},
+        Thank you for your generous donation of ${} to our cause
```
PythonCHB: you might want to use a format string so that you'll get 2 decimal places: $1,234.00

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+        Because of donors like you, we are able to execute our mission
+        to support at risk youth in achieving academic success.
+        Your contribution also ensures we will have the funds to develop
+        new programs to reach those most vulnerable.
+        
+        Sincerely,
+        Team Umizoomi'''.format(getPerson, getDonation))
+
+
+def getDonation():
```
PythonCHB: see pep8 -- this is a common naming style, but not so much in Python. PEP 8 suggests:

`get_donation`

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+    """
+    Asks user to specify donation amount and will error out if 
+    User specifies other than an int amount.
+    This could be modified in next version so entry can include 2 decimal places.
+    :Param: none
+    returns: donation retrieved from user
+    """
+    getDonation = ''
```
PythonCHB: it's good for function names to be verbs or action words, like "get_something", but this is not getting anything, so a noun is better -- why not `donation`?

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+    while not type(getDonation) is int:
+        try:
+            getDonation = int(input("Please enter donation amount:"
+                                    "==>"))
+        except ValueError:
+                print('Sorry, that isn\'t a valid integer. Please retry')
```
PythonCHB: you can use double quotes so you don't need to escape the apostrophe:

"Sorry, that isn't a valid integer. Please retry"

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+    return getDonation
+
+
+def thankyou():
+    """
+    :param:
+    :returns:
+    """
+    getPerson = ''
+    donation = ''
```
PythonCHB: any need to pre-define this?

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+    while getPerson not in donor_db:
```
PythonCHB: you should probably have some kind of "find_donor" function, rather than relying on a perfect match of name to key in a dict. Then you can then make that smarter in the future if you want.

PythonCHB: then you can put this kind of normalization in there:

getPerson.strip().lower().title()

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+        else:
+            try:
+                donation = getDonation()
+                donor_db[getPerson].append(donation)
+            except KeyError:
+                print("This new person will be added to the dictionary.")
+                donation = getDonation()
+                donor_db[getPerson] = [donation]
+            finally:
+                printThankYou(getPerson, donation)
```
PythonCHB: nicely done!

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+def printThankYou(getPerson, getDonation):
+    """
+    prints thank you letter for person and donation specified
+    :param: User provided person and donation amount
+    :returns: none (will return letter to save to disk shortly)
+    """
+    print(assemble_thank_you(getPerson, getDonation))
+
+
+def makereport():
```
PythonCHB: Consistent naming style -- PEP 8 recommends: `make_report` -- but you certainly don't want to mix that style with the "camelCase" style: `printThankYou`.

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+    """
+    Calculates values to put into report
+    :param:
+    :return:
+    """
+    donorReport = [[round(float(sum(value)), 2), key, len(value),
```
PythonCHB: nice comprehension -- though maybe a bit long.

perhaps define a function that computes the stuff the repot, and then call that in the comprehension.

And avoid names like "key" and "value" when they mean something, like "name" and "donations"

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+                    round(float(sum(value) / len(value)), 2)]
+                    for key, value in donor_db.items()]
+    sortedReport = sorted(donorReport)
+    ascendingReport = sortedReport[::-1]
```
PythonCHB: you can add `reverse=True` to the sorted() call, to get a reverse sort.

and if you don't need the unsorted version, you can call the list .sort() method to sort it in place.

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+    printReport(ascendingReport)
+
+
+def printReport(ascendingReport):
+    """
+    print report using f.strings
+    :param (name): sorted list in ascending order that
+                   contains required information
+                   for each donor
+    :return: It prints, neet to put into a return statement####
```
PythonCHB: yup!

I like to put comments in like:

```
# fixme: need to put into a return statement
```
then I can search for "fixme" to see stuff I need to do.

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+    produces a clean exit from the program
+    """
+    print("Exiting program...\n")
+    sys.exit()
+
+
+def main():
+    print("Welcome to the Mailroom!")
+    answer = ""
+    while answer != 'x' or 'r' or 't' or 'a':
```
PythonCHB: a little slicker way to spell this:

```
while answer not in ('x', 'r', 't',  'a'):
```
and I don't think that line means what you think it does!

```
In [9]: answer = 'x'                                                                           

In [10]: answer != 'x' or 'r' or 't' or 'a'                                                    
Out[10]: 'r'

In [11]: answer = 'r'                                                                          

In [12]: answer != 'x' or 'r' or 't' or 'a'                                                    
Out[12]: True
```
you would need:

```
not (answer == 'x' or answer == 'r' or answer == 't' or answer == 'a' )
```

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+        try:
+            print("\n\nPlease select from the following:")
+            print("x - Exit\nt - Thank you letter\n" +
+                  "a - send thank you letters to all donors\n" +
+                  "r - Create a Report\n")
+            answer = input(' ==> ')
+            answer = answer.strip()
+            answer = answer[0:1].lower()
+            user_choices.get(answer)()
+        except TypeError:
```
PythonCHB: nice! -- though it might be a bit more clear to do:

```
    user_choices[answer]()
except KeyError:
    retry()
```

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+    return None
+
+
+if __name__ == '__main__':
+    user_choices = {'t': thankyou,
+                    'r': makereport,
+                    'x': exitout,
+                    'a': alldonors}
+
+    donor_db = {"William Gates III": [653772.32, 12.17],
```
PythonCHB: careful here -- I think the III will get mangled by using .title() case -- so you may want to normalize your names (and dict keys) some other way.

One option is to store the name as original given, and then use a normalized version as the key in the dict.

PR #158, opened 2018-11-16
Slicing Lab

Misseswaddles:
I kept the Slicing lab on the master branch since I began the lab on master.


students/C_Farris/session03/slicing_lab.py
```diff
@@ -0,0 +1,98 @@
...
+
+    reverseTheList(myList)
+    reverseTheList(myString)
+
+    mixByThirds(myList)
+    mixByThirds(myString)
+    mixByThirds(myTest)
+    mixByThirds(myTest2)
+
+    assert swapfirstlast(tuple(myList)) == (13, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 1)
```
PythonCHB: These functions should return the same type as they are passed - so no need to convert to test. 

But this all looks great!

PR #162, opened 2018-11-20
submit Lesson 06 HW

ShujinCao:


students/JackieCao/Lesson05/except_exercise.py
```diff
@@ -0,0 +1,54 @@
...
+langs = ['java', 'c', 'python']
+
+try:
+    more_joke = more_fun(langs[0])
+except IndexError:
+    more_joke = more_fun(langs[1])
+else:
+    more_fun(langs[2])
+finally:
+    last_fun()
```
PythonCHB: perfect!

students/JackieCao/Lesson04/trigrams.py
```diff
@@ -0,0 +1,74 @@
+#!/usr/bin/env python3
+"""
+Trigrams
+"""
+import random
+import sys
+
+def read_book(filename):
+    with open(filename) as f:
+        lines = [line.rstrip('\n') for line in f]
```
PythonCHB: .strip() will strip all whitespace from both ends -- probably what you want.

students/JackieCao/Lesson04/trigrams.py
```diff
@@ -0,0 +1,74 @@
...
+    return lines
+
+def make_words(book, stat_line, end_line):
+    # specify the start and ending line (exclusive)
+    n = [i for i,x in enumerate(book) if x == start_line]
```
PythonCHB: this is looping through the book twice, and not really a list comp, as you don't actually want a list when you're done.

clever though.

students/JackieCao/Lesson04/trigrams.py
```diff
@@ -0,0 +1,74 @@
...
+    return lines
+
+def make_words(book, stat_line, end_line):
+    # specify the start and ending line (exclusive)
+    n = [i for i,x in enumerate(book) if x == start_line] 
+    N = [i for i,x in enumerate(book) if x == end_line]
+    text = list()
+    for i in range(n[0]+1,N[0]):
+        if book[i] != '':
+            add_word = book[i].split()
```
PythonCHB: careful with your names -- this is a bunch of words, not one.

students/JackieCao/Lesson04/trigrams.py
```diff
@@ -0,0 +1,74 @@
...
+        pair = (w1, w2)
+        if pair in tris:
+            tris[pair].append(w3)
+        else:
+            tris[pair] = [w3]
+    return tris
+
+def build_text(trigrams):
+    start_key = random.choice(list(trigrams.keys()))
+    l = list(start_key)
```
PythonCHB: what's "l" -- use meaningful names!

students/JackieCao/Lesson04/trigrams.py
```diff
@@ -0,0 +1,74 @@
...
+    while True:
+        # tell the loop when to stop
+        if len(l) == len(words):
```
PythonCHB: not necessarily want the same length -- and words is a global here -- it may not be available.

students/JackieCao/Lesson04/trigrams.py
```diff
@@ -0,0 +1,74 @@
...
+            break
+        elif start_key in trigrams:
+            the_third = random.choice(trigrams[start_key])
+            l.append(the_third)
+            # re-define the start_key
+            start_key = tuple(l[-2:])
```
PythonCHB: perfect logic here.

students/JackieCao/Lesson04/trigrams.py
```diff
@@ -0,0 +1,74 @@
...
+        else:
+            start_key = random.choice(list(trigrams.keys()))
+    new_text = " ".join(l)
+    if new_text[-1].isalpha:
+        new_text = new_text + '.'
```
PythonCHB: better to add that while still in the  list -- this is making a while new big string just to add the .

and only a . at the end of the while thing?

students/JackieCao/Lesson04/trigrams.py
```diff
@@ -0,0 +1,74 @@
...
+    else:
+        new_text[-1] = '.'
+    return new_text
+
+if __name__ == "__main__":
+    import argparse
+    parser = argparse.ArgumentParser()
```
PythonCHB: nice!

students/JackieCao/Lesson05/comprehension.py
```diff
@@ -0,0 +1,68 @@
...
+    l.append(set_n)
+print(l)
+
+# do it all as a one-liner
+l = [{s for s in range(21) if s % i == 0} for i in divisors]
+print(l)
+
+
+
+
```
PythonCHB: all good

PR #167, opened 2018-11-20
Mailroom 4

Misseswaddles:
Pull request for Mailroom 4 with test suite.
On vacation for the week so hoping to get this version in on time!


students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+        Because of donors like you, we are able to execute our mission
+        to support at risk youth in achieving academic success.
+        Your contribution also ensures we will have the funds to develop
+        new programs to reach those most vulnerable.
+        
+        Sincerely,
+        Team Umizoomi'''.format(getPerson, getDonation))
+
+
+def getDonation():
```
charlesrobison: All functions should be snake_case - looks like the ones above follow this convention, but then stopped from here down...

students/C_Farris/session05/Mailroom_3.py
```diff
@@ -0,0 +1,211 @@
...
+        else:
+            try:
+                donation = getDonation()
+                donor_db[getPerson].append(donation)
+            except KeyError:
+                print("This new person will be added to the dictionary.")
+                donation = getDonation()
+                donor_db[getPerson] = [donation]
+            finally:
+                printThankYou(getPerson, donation)
```
charlesrobison: Note that when a donor is not in the db, their donation amount is requested twice.  Can you identify why?

students/C_Farris/session06/Mailroom_4.py
```diff
@@ -0,0 +1,249 @@
...
+    """
+    return dedent(
+        '''\tDear {},
+        Thank you for your generous donation of ${} to our cause.
+        
+        Sincerely,
+        The Team'''.format(getPerson, getDonation))
+
+
+def getDonation(getDonation=''):
```
charlesrobison: All function names camel_case

students/C_Farris/session06/Mailroom_4.py
```diff
@@ -0,0 +1,249 @@
...
+if __name__ == '__main__':
+    user_choices = {'t': thankyou,
+                    'r': makereport,
+                    'x': exitout,
+                    'a': alldonors}
+
+
+    prepare_to_write_to_disk()
+
+    main()
```
charlesrobison: This works really well!

PR #169, opened 2018-11-21
Submitting Week 6 homework

kmonster462:
Adding mailroom_4 and test file, added a few changes to mailroom_3 when I found some bugs!


PythonCHB:
not sure what what messed up -- but this looks all good :-)

students/Kate_Koehn/Session06/mailroom_4.py
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env python3
+
+"""
+mailroom assignment
+"""
+
+# Mailroom Part 3
+
+# import relevant dictionaries
```
PythonCHB: you are importing "Modules" or "Packages".

students/Kate_Koehn/Session06/mailroom_4.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Cheers,
+        Kate Koehn
+    '''.format(donor_name, donation_amount))
+
+
+def get_donor_key(donor_name):
+    """looks up key in donor dict"""
+    donor_key = donor_name.strip()
+    for key in donors.keys():
```
PythonCHB: the whole point of using a dict is that you can do a direct lookup -- this should be something like:

```
def get_donor(donor_name):
    donor_key = donor_name.strip()
    try:    
        return  donors['donor_key']
    except KeyError:
        return None
```

PythonCHB: If it's well tested, you should be able to simplify the code and still pass tests :-)

students/Kate_Koehn/Session06/mailroom_4.py
```diff
@@ -0,0 +1,172 @@
...
+        if donor_key == key:
+            return True
+    return False
+
+
+def add_new_donor(donor_name):
+    name = donor_name.strip()
+    if name not in donors.keys():
```
PythonCHB: if name not in donors():

some thing -- faster to run!

students/Kate_Koehn/Session06/mailroom_4.py
```diff
@@ -0,0 +1,172 @@
...
+        donors[name] = []
+    return donor_name
+
+
+def add_donation_to_donor(donor_name, donation):
+    donors[donor_name].append(donation)
+
+
+def check_donation(donation):
+    donation = float(donation)
```
PythonCHB: this could use a docstring (they all could :-) )

"""
checks if a string for a donation is a valid float.

:param donation: the donation string

:returns: a float with the value.

Raises a ValueError if donation is not a valid value
"""

students/Kate_Koehn/Session06/mailroom_4.py
```diff
@@ -0,0 +1,172 @@
...
+    if math.isnan(donation) or math.isinf(donation) or round(donation, 2) <= 0.00:
+        raise ValueError
+    return donation
+
+
+def record_new_donation(donor_name):
```
PythonCHB: nice! virtually no logic in here :-)

students/Kate_Koehn/Session06/mailroom_4.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+def thanks():
+    """sub-menu to record a new donation and thank the user for their donation"""
+    while True:
+        print("\n\n~ Thanking Your Donors ~\n")
+        donor_name = input("Enter a donor's name or 'list' to see all donors or 'menu' to exit to main menu: ").strip()
+
+        if donor_name == "list":
+            for name in donors:
```
PythonCHB: seems trivial, but s "list_donors" function may be called for -- particular if you might want to make a fancier listing at some point.

students/Kate_Koehn/Session06/mailroom_4.py
```diff
@@ -0,0 +1,172 @@
...
+        print("Your letter has been written to:", donor)
+        filename = os.path.join(OUT_PATH, filename)
+        open(filename, 'w').write(letter)
+
+
+if __name__ == "__main__":
+
+    donors = get_sample_donors()
+
+    prepare_to_run()
```
PythonCHB: maybe "prepare_to_run" could call get_sample_donors() ?

though you need to think about how to handle the global namespace..

students/Kate_Koehn/Session06/mailroom_test.py
```diff
@@ -0,0 +1,92 @@
...
+def get_sample_donors():
+    donors = {"Rick Sanchez": [3.00, 1.00],
+              "Liz Lemon": [4000.00, 3000.00, 6000.00],
+              "Andy Dwyer": [10.00],
+              "Brendan Small": [3.00, 10.00],
+              "Coach McGuirk": [2.00, 1.00]}
+    return donors
+
+
+mr.donors = get_sample_donors()
```
PythonCHB: nice!

students/Kate_Koehn/Session06/mailroom_test.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def test_adds_new_donor():
+    mr.add_new_donor("Jake Ham")
+
+    assert "Jake Ham" in mr.donors
```
PythonCHB: I think you have a function that checks if a name is in the db, yes? then good to use that, in case the structure of the db changes.

students/Kate_Koehn/Session06/mailroom_test.py
```diff
@@ -0,0 +1,92 @@
...
+    assert [] == mr.donors.get("Jake Ham")
+
+
+
+def test_add_new_donation():
+    mr.add_donation_to_donor("Jake Ham", 500)
+
+    assert 500 in mr.donors.get("Jake Ham")
```
PythonCHB: how do you know that there wasn't already a 500 in there?

PR #172, opened 2018-11-23
Session06 for grading

ktbzzz:
To be graded:
Session06 - ready for grading
Session07 - ready for grading
Session08 - ready for grading


.idea/modules.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
ktbzzz: I have removed this folder on the latest commit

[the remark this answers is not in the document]
ktbzzz: I have removed the .idea folder

ktbzzz:
I have removed the .idea folder now, thank you

PythonCHB:
Needs review of 6,7,8

students/JaeKim/session08/test_circle.py
```diff
@@ -0,0 +1,111 @@
...
+
+    assert s.volume == 904.7786842338603
+
+
+def test_sphere_area():
+    s = Sphere(2)
+    s2 = Sphere(3)
+
+    assert s.area == 50.26548245743669
+    assert s2.area == 113.09733552923255
```
PythonCHB: careful here -- floating point calculations can be no-quite-perfect.

check out `math.isclose()` for a better way to handle FP comparisons.

students/JaeKim/session08/test_circle.py
```diff
@@ -0,0 +1,111 @@
...
+    d = Circle(2)
+
+    assert (c + d) == 'Circle(4)'
+
+
+def test_mul_():
+    c = Circle(2)
+    d = Circle(2)
+
+    assert (c + d) == 'Circle(4)'
```
PythonCHB: this should not be creating a string!

add two circles, get a string ??

students/JaeKim/session08/circle.py
```diff
@@ -0,0 +1,55 @@
...
+    def area(self):
+        return pi * self.radius**2
+
+    def __str__(self):
+        return f'Circle with radius: {self.radius:.6f}'
+
+    def __repr__(self):
+        return f'{self.__class__.__name__}({self.radius})'
+
+    def __add__(self, other):
```
PythonCHB: these should be returning new Circle objects -- not strings!

students/JaeKim/session08/circle.py
```diff
@@ -0,0 +1,55 @@
...
+        return f'{self.__class__.__name__}({self.radius + other.radius})'
+
+    def __mul__(self, other):
+        return f'{self.__class__.__name__}({self.radius * other.radius})'
+
+    def __gt__(self, other):
+        return {self.radius > other.radius}
```
PythonCHB: the curly brackets create a set object -- so rather than returning True or False, you are getting a set with `True` or `False` in it.

Which is a non-empty set, and therefor "Truthy" -- are you testing these? do the tests pass???

students/JaeKim/session08/circle.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __eq__(self, other):
+        return {self.radius == other.radius}
+
+
+class Sphere(Circle):
+    def __str__(self):
+        return f'Sphere with radius: {self.radius:.6f}'
+
+    def __repr__(self):
```
PythonCHB: using __class__ here is really great -- why not also int the __str__ ?

Also, if you use __class__ here, then you can put it in the base class (Circle) and it will "just work" here!

students/JaeKim/session08/test_circle.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def test_gt_():
```
PythonCHB: you need to make sure that it returns FAlse when it should as well:

`assert not c > c`

I think that test would show an error in your code!

students/JaeKim/session08/test_circle.py
```diff
@@ -0,0 +1,111 @@
...
+    d = Circle(4)
+
+    assert d > c
+
+
+def test_eq_():
+    c = Circle(2)
+    d = Circle(2)
+
+    assert c == d
```
PythonCHB: also test that you get false if circles are not equal!

students/JaeKim/session08/test_circle.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def test_sorted():
+    sorted_circles = [Circle(5), Circle(2), Circle(10)]
+    sorted_circles.sort()
+
+    assert sorted_circles[0] == Circle(2)
+    assert sorted_circles[1] == Circle(5)
+    assert sorted_circles[2] == Circle(10)
```
PythonCHB: I think there are errors in your comparisons -- I wonder if you have a larger collection of circles to sort, if they would show up?

students/JaeKim/session06/mailroom.py
```diff
@@ -0,0 +1,127 @@
...
+            return
+        add_donor(name, donation)
+    elif name in players:
+        send_thank_you(name)
+
+
+def send_all_donors():
+    for player in players:
+        donation_sum = sum(players[player])
+        email_path = tempfile.gettempdir() + '\\' + player + ".txt"
```
charlesrobison: This is an issue for a non-Windows environment.  Recommend a solution that works for any platform.

students/JaeKim/session06/mailroom.py
```diff
@@ -0,0 +1,127 @@
...
+"-------------------------------------------------------------------------")
+
+    format = "%-20s  $ % 16f% 14i  $% 16f"
+
+    for player in players:
+        report += format % (player, sum(players[player]), len(players[player]), (sum(players[player]) / len(players[player])))
+
+    print(report)
+
+    return report
```
charlesrobison: This is lining up in an odd way for me.

students/JaeKim/session06/mailroom.py
```diff
@@ -0,0 +1,127 @@
...
+                return
+            thank_you(name)
+        elif answer == '2':
+            make_report()
+        elif answer == '3':
+            send_all_donors()
+
+
+if __name__ == "__main__":
+    main()
```
charlesrobison: Overall works well - just a couple suggestions.

students/JaeKim/session06/test_mailroom.py
```diff
@@ -0,0 +1,38 @@
...
+    assert "Carmelo Anthony" in report
+    assert "Kevin Durant" in report
+    assert "Michael Jordan" in report
+
+
+def test_make_report_with_new_donor():
+    mail.add_donor("Jae", 500)
+    report = mail.make_report()
+
+    assert "Jae" in report
```
charlesrobison: Looks good!

students/JaeKim/session07/html_render.py
```diff
@@ -0,0 +1,149 @@
...
+        out_file.write("<!DOCTYPE html>\n")
+        super().render(out_file, cur_ind)
+
+
+class A(OneLineTag):
+    tag = 'a'
+
+    def __init__(self, link, content=None, **kwargs):
+        kwargs['href'] = link
+        super().__init__(content, **kwargs)
```
charlesrobison: Looks like this isn't complete.  Runs only through step 6 in run_html_render.py.  Recommend completing those steps!

students/JaeKim/session08/circle.py
```diff
@@ -0,0 +1,55 @@
...
+    @property
+    def volume(self):
+        radius = self.radius
+        volume = 4.0 / 3.0 * pi * radius ** 3
+
+        return volume
+
+    @property
+    def area(self):
+        return 4 * pi * self.radius * self.radius
```
charlesrobison: Looks great!

PR #180, opened 2018-11-27
HTML Render

rushinstuffin:
First attempt at this. Still work to be done. Definitely need to keep exploring it.


students/KyleBarry/session06/test_mailroom_04.py
```diff
@@ -0,0 +1,39 @@
...
+def test_make_list():
+    result = mr.make_list()
+
+    assert len(result) == len(mr.donors)
+
+
+def test_add_donation():
+    result = mr.add_donation("Tara Towers", 9999)
+
+    assert result[-1] == 9999
```
PythonCHB: Better to also check if the db itself has been altered appropriately.

students/KyleBarry/session06/test_mailroom_04.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_add_donation_new():
+    result = mr.add_donation_new("Jane Janeson", 1000)
+
+    assert result in mr.donors.values()
```
PythonCHB: careful about tests (and code)  like this -- they are closely tied to the data structure you are using.

Do you have a "find_donor" or similar function? You should.

students/KyleBarry/session06/test_mailroom_04.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_make_thank_you_file():
+    result = mr.make_thank_you_file("Kevin Smith", 10000)
+
+    # Just open(result) works, but this makes it clearer I think
+    assert bool(open(result)) is True
```
PythonCHB: or use `os.path.isfile()`

students/KyleBarry/session06/test_mailroom_04.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def test_make_report():
+    result = mr.make_report()
+
+    # Just open(result) works, but this makes it clearer I think
+    assert bool(open(result)) is True
```
PythonCHB: This checks that a file was made -- good start. But is the contents of the report correct?

students/KyleBarry/session07/html_render.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class Anchor(OneLineTag):
+
+    tag = "a"
+
+    def __init__(self, link, content=None, **kwargs):
+        kwargs['href'] = link
+        super().__init__(content, **kwargs)
+
```
PythonCHB: good so far!

students/KyleBarry/session08/circle.py
```diff
@@ -0,0 +1,34 @@
...
+    # Figure out how to make this diameter settable.
+    @property
+    def diameter(self):
+        return self.radius*2
+
+    @property
+    def area(self):
+        return round(self.radius**2*math.pi, 2)
+
+
```
PythonCHB: getting there -- all good so far.

students/KyleBarry/session08/sparse_array.py
```diff
@@ -0,0 +1,44 @@
...
+    def __getitem__(self, pos):
+        """Allow for accessing list by index and slicing"""
+        return self.seq[pos]
+
+    def append(self, value):
+        """Append to end of array"""
+        self.seq.append(value)
+        return self.seq
+
+    # Was not able to figure out how to make del [] work
```
PythonCHB: you need to define a ``__delitem__`` magic method.

students/KyleBarry/session08/sparse_array.py
```diff
@@ -0,0 +1,44 @@
+"""Emulate a list in a custom class"""
+
+
+class SparseArray:
+    """Create array objects that act as lists"""
+
+    def __init__(self, seq):
+        """Establish variable for sequence argument"""
+        self.seq = seq
```
PythonCHB: this isn't going to give you any sparse behaviour -- you need to store the non-zero values some other way.

Also if you want to store stuff in a sequence, you probaby want to know exactly what type it is here -- for instance if someone init's with a tuple, then any of the mutating methods wouldn't work. So at least:

`self.seq = list(seq)`

which isn't a horrible way to develop and test you magic methods, and then you can make it spase later -- but he sparse part is the tricky part :-)

PythonCHB:
you can do a new PR when you're done iwth html_render or Circle.

Do let us know when something is ready for review when you do a PR.

(or you have questions)

PR #182, opened 2018-11-28
Completed str formatting lab and mailroom

fyfynguyen:
A few questions about mailroom.py...
- _**Lines 28-36**_: can I use """ triple quotes here? 
- **_Lines 50-57_**: originally kept getting "'tuple' object has no attribute 'append'" which made sense but still a little unclear why the donor object can be manipulated in this instance (after the database structure is defined for new users). The mutable inside the immutable is tripping me up.
- **_PEP8 style_**: a little unclear how to wrap long lines so they meet 79 char limit based on the documentation -- _"The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses."_ Is what I did at lines 41-42, 66-67, 72-74 okay? My linter didn't yell at me...


PythonCHB:
thanks for the questions -- see the inline notes for answers.

students/FyfyNguyen/session03/mailroom.py
```diff
@@ -1,35 +1,103 @@
...
+def find_donor(name_entered):
+    for donor in donors:
+        if name_entered.strip().lower() == donor[0].lower():
+            return donor
+    return None
+
+
+def send_email(name_entered, donation):
+    print(
+        """Dear {},
```
PythonCHB: yes, tripple quotes is perfect for this sort of thing.

But as you noticed, it does require to to not indent, which is a bit ugly.

The the `dedent` function in the `textwrap` module can help here:

https://docs.python.org/3.1/library/textwrap.html#textwrap.dedent

students/FyfyNguyen/session03/mailroom.py
```diff
@@ -1,35 +1,103 @@
...
+        if name_entered == "list":
+            print_donors()
+        else:
+            break
+
+    donation = float(input("Enter donation amount >>> "))
+    donor = find_donor(name_entered)
+    if donor is None:
+        donor = (name_entered.title(), [])
+        donor[1].append(donation)
```
PythonCHB: you have it exactly right: this is a mutable inside an immutable.

A tuple being immutable means you can't change its contents. -- that is list of donations has to be the SAME list -- but you can still change that list in place. what you can't do is replace it with a new list.

students/FyfyNguyen/session03/mailroom.py
```diff
@@ -1,35 +1,103 @@
...
+    )
 
-donors = [("Fred Jones", [100, 200, 300]),
-        ("Amy Schumer", [2000, 4000, 1000])
-        ]
 
 def thank_you():
-    print("do the thank you think now")
+    while True:
+        name_entered = input("Enter the name of a donor or 'list' to view "
```
PythonCHB: yes, a great way to make the lines short enough for PEP8

students/FyfyNguyen/session03/mailroom.py
```diff
@@ -1,35 +1,103 @@
...
+
+    send_email(name_entered, donation)
+
+
+def sort_key(donors):
+    return donors[1]
+
+
+def create_report():
+    print("{:30s} | {:11s} | {:9s} | {:12s}".format(
```
PythonCHB: also a good way to wrap lines, though I personally prefer this style:

```
    print("{:30s} | {:11s} | {:9s} | {:12s}".format("Donor Name",
                                                    "Total Given",
                                                    "Num Gifts",
                                                    "Average Gift"))
```

that is -- if there are a lot of arguments or parameters, put them each on one line -- I find it more readable, but ti's personal preference.

students/FyfyNguyen/session03/slicing_lab.py
```diff
@@ -1,8 +1,16 @@
 #!/usr/bin/env python3
 
 """
+<<<<<<< HEAD
```
PythonCHB: you got a bunch of merge conflicts here -- mostly jsut extra / missing newlines. not sure whay that is, but I think it's less liley to happen (and/or be easier to fix) if you do:

`git pull upstream master`

before pushing changes to your gitHub repo.

PythonCHB: But this has already been reviewed, so you can ignore it, too :-)

students/FyfyNguyen/session03/strformat_lab.py
```diff
@@ -0,0 +1,54 @@
...
+print(f"file_{tup[0]:03d}: {tup[1]:.2f}, {tup[2]:.2e}, {tup[3]:.3e}")
+
+
+# Task Three
+# Format arbitrary object just using {}
+# Multiply a list by length of the sequence to get number of "{}" needed
+# Use str.join() method to join seq separated by commas
+def formatter(seq):
+    seq_len = len(seq)
+    string = ("the {} numbers are: " + ", ".join(["{}"]*seq_len)).format(
```
PythonCHB: nicely done!

students/FyfyNguyen/session03/strformat_lab.py
```diff
@@ -0,0 +1,54 @@
...
+# Task Six
+data = [("Beyonce", 37, 355.00),
+        ("Jay-Z", 48, 810.00),
+        ("Blue Ivy", 6, 1, 165.00)
+        ]
+
+print("{:20s} | {:5s} | {:8s}".format("Name", "Age", "Price"))
+print("-" * 42)
+for item in data:
+    print("{:20s}   {:5d}   $ {:8,.2f}".format(*item))
```
PythonCHB: nice work -- all good :-)

PR #183, opened 2018-11-28
Lesson 08 Circle class and test

Misseswaddles:


students/C_Farris/session08/circle.py
```diff
@@ -38,7 +38,6 @@ def __add__(self, other):
 
     def __mul__(self, other):
         if self.radius:
-            print("here!")
             return f'{self.__class__.__name__}({(self.radius * other)})'
```
PythonCHB: you want these all to return Circle objects -- not strings!

Misseswaddles:
Yikes!

I will fix and resubmit!

Best,
Carol

PythonCHB:
closing now -- you can re-open when it's ready.

PR #185, opened 2018-12-01
Incomplete html_render

vanulor:
I can't get indent working properly.  I would be willing to do a code review in class.


PythonCHB:
thanks for the offer -- we just may do that!

meanwhile, see my comments, maybe you'll figure it out :-)

students/BrianNisonger/session7/html_render.py
```diff
@@ -1,18 +1,177 @@
...
+            self.content = [content]
+        self.kwargs = kwargs
 
     def append(self, new_content):
-        pass
+        self.content.append(new_content)
+
+    def _open_tag(self, indent=indent):
+        open_tag = f'<{self.tag}'
+        for attributes in self.kwargs:
```
PythonCHB: since you want both the keys and values:

```
for attr, value in self.wkargs.items():
    ...
```

students/BrianNisonger/session7/html_render.py
```diff
@@ -1,18 +1,177 @@
...
+            open_tag = open_tag + (
+                f' {attributes}="{self.kwargs[attributes]}"')
+        open_tag = open_tag + (">\n")
```
PythonCHB: no need for parens around a string.

students/BrianNisonger/session7/html_render.py
```diff
@@ -1,18 +1,177 @@
...
+
+    def render(self, out_file, indent=indent):
+        self.indent = indent
+        tag_indent = self.indent
+        out_file.write(tag_indent)
+        out_file.write(self._open_tag())
+        for content in self.content:
+            if content != None:
+                try:
+                    content.render(out_file, indent="")
```
PythonCHB: this is where you want to pass the current level of indentation on to the sub-elements.

students/BrianNisonger/session7/html_render.py
```diff
@@ -1,18 +1,177 @@
...
+                except AttributeError:
+                    out_file.write(content.strip())
```
PythonCHB: and you want to use the level of indentation here, too.

students/BrianNisonger/session7/html_render.py
```diff
@@ -1,18 +1,177 @@
...
+class P(Element):
+    tag = "p"
+
+
+class Head(Element):
+    tag = "head"
+
+
+class OneLineTag(Element):
+    indent = ""
```
PythonCHB: a one line tag could be indented -- you don't want to override that here!

students/BrianNisonger/session7/html_render.py
```diff
@@ -1,18 +1,177 @@
...
+
+    def append(self, new_content):
```
PythonCHB: does this need a custom append?

students/BrianNisonger/session7/html_render.py
```diff
@@ -1,18 +1,177 @@
...
+        self.content.append(f' {new_content}')
+
+    def _open_tag(self):
```
PythonCHB: DRY!

(don't repeat yourself)

the whole point of defining this as a method on the base Element class is that you can use it everywhere else.

students/BrianNisonger/session7/html_render.py
```diff
@@ -1,18 +1,177 @@
...
+
+class SelfClosingTag(Element):
+    indent = ""
+
+    def __init__(self, content=None, **kwargs):
+        if content is not None:
+            raise TypeError("SelfClosingTag can not contain any content")
+        super().__init__(content=content, **kwargs)
+
+    def _open_tag(self):
```
PythonCHB: same here -- only override the methods that need to be different.

students/BrianNisonger/session7/html_render.py
```diff
@@ -1,18 +1,177 @@
...
+class Meta(SelfClosingTag):
+    tag = "meta"
+    indent = ""
+
+
+class A(OneLineTag):
+    tag = "a"
+
+    def __init__(self, link, content=None, **kwargs):
+        kwargs['href'] = link
```
PythonCHB: yup!

students/BrianNisonger/session7/html_render.py
```diff
@@ -1,18 +1,177 @@
...
+class H(OneLineTag):
+    tag = "h"
+
+    def __init__(self, level, content=None, **kwargs):
+        super().__init__(content, **kwargs)
+        self.level = level
+
+    def _open_tag(self):
+        open_tag = f'<{self.tag}{self.level}'
+        for attributes in self.kwargs:
```
PythonCHB: is there a way to re-use this code?

PythonCHB:
merging 'cause I've reviewed it.

This may be a good example to review in class --thanks!

PR #186, opened 2018-12-04
mainly for session07 final

habtamua:
- session07 final update
- not finished yet session08
- session04 there was merge issues for mailroom


students/HABTAMU/session07/test_html_output7.html
```diff
@@ -0,0 +1,20 @@
+<!DOCTYPE html>
```
PythonCHB: you generally don't want to put output like this into git.

always check carefully what is being added when you commit -- do a "git status" first!

students/HABTAMU/session07/html_render.py
```diff
@@ -0,0 +1,127 @@
...
+        self.tag = self.tag + str(level)
+        super(H,self).__init__(content,**kwargs)
+
+class Ul(Element):
+    tag = 'UL'
+
+class Li(OneLineTag):
+    tag = 'li'
+    def append(self, content):
+        if isinstance(content, OneLineTag) or type(content) is str:
```
PythonCHB: do you want to be able to put a OneLIneTag inside an Li element? I"m not HTML expect, but that seems weird to me.

students/HABTAMU/session07/html_render.py
```diff
@@ -0,0 +1,127 @@
...
+    #     if isinstance(content, OneLineTag) or type(content) is str:
+    #         self.contents.append(content)
+    #     else:
+    #         raise ValueError(f"You can only append an instance of OneLineTag or a string to tag '{self.tag}'.")
+
+    def append(self, content):
+        raise NotImplementedError
+
+    def render(self, out_file, ind=""):
+        attribute_str = "".join([f' {attribute}="{value}"' for attribute, value in self.html_attributes.items()])
```
PythonCHB: this is a lot of repeated code from the superclass `render` -- can you refactor that out?

students/HABTAMU/session07/html_render.py
```diff
@@ -0,0 +1,127 @@
...
+
+class Head(Element):
+    tag = 'head'
+
+class SelfClosingTag(Element):
+    # tag = 'hr'
+    
+    def render(self, out_file, ind=""):
+        # render self closing tags with attributes to out_file
+        attribute_str = "".join([f' {attribute}="{value}"' for attribute, value in self.html_attributes.items()])
```
PythonCHB: repeated code here, too....

students/HABTAMU/session07/html_render.py
```diff
@@ -0,0 +1,127 @@
...
+        else:
+            self.contents = [content]
+        # print("contents is:", self.contents)
+
+
+    def append(self, new_content):
+        self.contents.append(new_content)
+        
+    def render(self, out_file, ind=""):
+        # loop through the list of contents
```
PythonCHB: this comment a bit out of place now...

students/HABTAMU/session07/html_render.py
```diff
@@ -0,0 +1,127 @@
...
+        attribute_str = "".join([f' {attribute}="{value}"' for attribute,value in self.html_attributes.items()])
+        out_file.write(ind + "<{}{}>\n".format(self.tag,attribute_str))
+        for content in self.contents:
+            # Duck Typing on the Fly
+            try:
+                content.render(out_file, ind + "  ")
```
PythonCHB: don't hard-code that indentation! 

It should be a class attribute on the Element class

students/HABTAMU/session07/html_render.py
```diff
@@ -0,0 +1,127 @@
...
+A class-based system for rendering html.
+"""
+
+# This is the framework for the base class
+class Element(object):
+
+    tag = "html"
+    indent = "  "
+    # html_attributes = {"pid": "intro", "style": "text-align: center", "width":"400"}
+    html_attributes = dict()
```
PythonCHB: why a class attribute for this?

it'll get overwritten by the __init__ anyway.

students/HABTAMU/session06/mailroom_4.py
```diff
@@ -0,0 +1,112 @@
...
+    generates_letter(d_name, contribution)
+
+
+def create_report():
+    # create a report for donors contribution as a table formate.
+    """prints list of donors, sorted by total historical donation amount."""
+    print("{:<20}| {:<10} | {:<10} | {:<12}".format("Donor Name", "Total Given", " Num Gifts", "Average Gift"))
+    print("-" * 60)
+    # list comprehension here
+    [print("{:<21} ${:<15.2f} {:<10} ${:<12.2f}".format(
```
PythonCHB: same as above -- you can generate the rows with a list comp, but don't print them from there.

students/HABTAMU/session06/mailroom_4.py
```diff
@@ -0,0 +1,112 @@
...
+
+def save_letter_todisk():
+    """ Save generates thank you letter, 
+       and writes each letter to disk as a text file. 
+    """
+    for donor,contributions in donors_info.items():
+        with open(f"{donor.replace(' ', '_')}.txt","w") as output:
+            output.write(
+                """
+Dear {},
```
PythonCHB: you want to generate the letter separately -- then use the same code whether you are writing to disk, or printing to the screen, or ....

students/HABTAMU/session06/test_mailroom_4.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python3
+
+"""
+unit tests for the mailroom program
+"""
+import os
+import sys
+
+import mailroom_4 as mailroom
+    """
```
PythonCHB: this should be a docstring at the top of the file, or comments

students/HABTAMU/session06/test_mailroom_4.py
```diff
@@ -0,0 +1,48 @@
...
+    # Test each donors and contrubution from saved letter. 
+    mailroom.save_letter_todisk()
+    for donor, contribution in {"Bill Gates": "$16,500.00", "Paul Alen": "$10,600.00", "Mark Zuckerberg":"$42,082.11"}.items():
+        with open(f"{donor.replace(' ', '_')}.txt", "r") as txt:
+            content = txt.read()
+            assert donor in content
+            assert contribution in content
+
+if __name__ == '__main__':
+    test_generates_letter(capsys)
```
PythonCHB: you really want your tests to be runnable with pytest.

and where does capsys come from?

PythonCHB: Does this code even run? I'd expect an error -- with "capsys" not being defined.

students/HABTAMU/session07/html_render.py
habtamua: I assumed Li is one of OneLineTag tag and that works for me, or render as it suppose to be like this.

` <UL id="TheList" style="line-height:200%">
      <li>The first item in a list</li>
      <li style="color: red">This is the second item</li>
      <li>And this is a <a href="http://google.com">link</a>to google</li>
    </UL>
 `

students/HABTAMU/session07/html_render.py
habtamua: I just correct that.

students/HABTAMU/session06/test_mailroom_4.py
habtamua: `test_mailroom_4.py` actually run, but my bad it shouldn't be add as an object on test_generates_letter method. anyways I use capsys for testing purpose and its a fixtures allow access to capture the stdout/stderr output.

I commented out and still working.
`# if __name__ == '__main__':
#     test_generates_letter(capsys)
#     test_create_report(capsys)
#     test_save_letter_todisk()`

students/HABTAMU/session07/test_html_output7.html
PythonCHB: you want that DOCTYPE tag, yes,

but you don't want to put generated files, like `test_html_output7.html` in git.

students/HABTAMU/session06/test_mailroom_4.py
habtamua: Thank you! and here where i got from [https://docs.pytest.org/en/latest/capture.html](url)

PR #189, opened 2018-12-04
Finished circle assignment

vanulor:
I was pleasantly surprised by how the sort function just worked once I had the gt,lt,eq working, which was really cool.  Like they say sometimes Python just works like you think it should.


students/BrianNisonger/Session8/circle.py
```diff
@@ -0,0 +1,85 @@
+import math
+
+
+class circle():
+    def __init__(self, radius):
+        self._radius = radius
+
+    @property
```
PythonCHB: no need for properties if you're not going to do anything in them. 

just store self.radius as a regular attribute.

students/BrianNisonger/Session8/circle.py
```diff
@@ -0,0 +1,85 @@
...
+    def radius(self):
+        return self._radius
+
+    @radius.setter
+    def radius(self, radius):
+        self._radius = radius
+
+    @property
```
PythonCHB: these we have because we don't want to store diameter or area as attributes.

students/BrianNisonger/Session8/circle.py
```diff
@@ -0,0 +1,85 @@
...
+    def area(self):
+        return math.pi * (self._radius**2)
+
+    @area.setter
+    def area(self, value):
+        self._radius = math.sqrt(value / math.pi)
+
+    @classmethod
+    def from_diameter(circle_object, value):
```
PythonCHB: we usually use a generic name here, like "cls" or "klass" (can't use "class", it's reserved).

while this will be the Circle class in this case, if you subclass it, it's not longer a Circle -- it will be whatever the subclass is -- a Sphere, for instance.

students/BrianNisonger/Session8/circle.py
```diff
@@ -0,0 +1,85 @@
...
+        return (self.radius > other.radius)
+
+    def __lt__(self, other):
+        return (self.radius < other.radius)
+
+    def __eq__(self, other):
+        return (self.radius == other.radius)
+
+
+class sphere(circle):
```
PythonCHB: use Capital names for classes: Circle, Sphere -- PEP8.

students/BrianNisonger/Session8/test_Circle.py
```diff
@@ -0,0 +1,114 @@
...
+
+def test_change_radius_diameter():
+    c = circle(10)
+    c.diameter = 100
+    assert c.radius == 50
+
+
+def test_set_area():
+    c = circle(10)
+    assert math.isclose(c.area, (math.pi * 10**2), rel_tol=.01)
```
PythonCHB: nice!

though a relative tolerance of 0.01 is really huge! essentially only 2 significant figures -- this should easily support the default (which I think is 1e-8) and probably much more: try 1e-14 or so.

students/BrianNisonger/Session8/test_Circle.py
```diff
@@ -0,0 +1,114 @@
...
+
+def test_str():
+    c = circle(10)
+    assert str(c) == 'Circle with radius: 10.000000'
+    assert repr(c) == "Circle(10)"
+
+
+def test_math():
+    c1 = circle(10)
+    c2 = circle(10)
```
PythonCHB: a good idea to not use the same value for both in a test -- if your code was using just one radius, it would still pass!

students/BrianNisonger/Session8/test_Circle.py
```diff
@@ -0,0 +1,114 @@
...
+    assert c1 + c2 == 'Circle(20)'
+    assert c1 * 2 == 'Circle(20)'
+    assert 2 * c1 == 'Circle(20)'
+
+
+def test_relationships():
```
PythonCHB: this is a few too many for one test -- if the first one fails, the others wont run.

students/BrianNisonger/Session8/test_Circle.py
```diff
@@ -0,0 +1,114 @@
...
+
+def test_sphere():
+    s1 = sphere(10)
+    assert str(s1) == 'Sphere with radius: 10.000000'
+    assert repr(s1) == 'Sphere(10)'
+    assert math.isclose(s1.volume, 4188.79, rel_tol=.01)
+    assert math.isclose(s1.area, 1256.64, rel_tol=.01)
+    s1.volume = 10
+    assert math.isclose(s1.volume, 10, rel_tol=.01)
+    assert math.isclose(s1.radius, 1.34, rel_tol=.01)
```
PythonCHB: you may want to set rel_tol as a global in the test file, so you can change it everywere easily.

PR #190, opened 2018-12-04
Added Circle and Circle_Test Exercise for Session08

Zach-Cooper:
Thanks!!!


Zach-Cooper:
@charlesrobison - Hi Charles, how can I delete those files that are not mine. Not sure how to do this and I dont want to make any changes that might screw somthing up. Really confused what actually happened since I only pushed my assignments but it looks like other material showed up?? Im puzzled on this haha Thanks for the help.

PR #193, opened 2018-12-05
Submitting Session 8 HW

kmonster462:


students/Kate_Koehn/Session09/mailroom_oo/donor_modules.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @property
+    def avg_donations(self):
+
+
+# create properties for sum, avg, number donations
+#have class that manages the collections
+
+
+##haven't been able to test this yet.
```
PythonCHB: need to test as soon as you write any code!

students/Kate_Koehn/Session08/class_circle.py
```diff
@@ -0,0 +1,73 @@
...
+    def __lt__(self, other):
+        return self.radius < other.radius
+
+
+    def __gt__(self, other):
+        return self.radius > other.radius
+
+
+    def sort(self):
+        return self.radius.sort()
```
PythonCHB: did you try this??? I think it would give an error -- the radius is a number, it does not have a sort method!

-CHB

students/Kate_Koehn/Session08/class_circle.py
```diff
@@ -0,0 +1,73 @@
...
+
+
+class Sphere(Circle):
+
+    def __init__(self, radius):
```
PythonCHB: non need to write an __init__ if all it's going to do is call super().__init__ -- that will get called be default if you don't define one here.

PythonCHB:
Do finish OO mailroom if you can -- good to get a handle on OO, and testing !

PR #195, opened 2018-12-05
mailroom4 and Trig Complete

thanhnhan7338:
Can you please review if I submit homework correctly? Thanks!


students/ThanhNhan/lesson06/mailroom4.py
```diff
@@ -0,0 +1,251 @@
...
+            print(filename)
+            filename = os.path.join(DIR_NAME, filename)
+            open(filename, 'w').write(letter)
+        except FileNotFoundError:
+            print("couldn't open missing.txt")
+        finally:
+            pass
+
+
+def default():
```
PythonCHB: this is the default in the switch-dict, but give it a name for whatit represents: maybe "invalid_option" or something.

students/ThanhNhan/lesson06/mailroom4.py
```diff
@@ -0,0 +1,251 @@
...
+        n_time = len(values)
+        average = total_cost/n_time
+        report_donor.append([name, round(total_cost, 2), n_time, round(average, 2)])
+
+    print (report_donor)
+    #sort the report
+    report = sorted(report_donor, key=str_lname, reverse=False)
+
+    print("{:20}|  {:<10s}|  {:<10s}|  {:<10s} ".format(*title))
+    print("".join(["-"] * 62))
```
PythonCHB: you don't want to print here -- have this function return a big string, or a list of rows in the report.

then print them in another function -- that way you can test this, and re-use it with a different interface.

students/ThanhNhan/lesson06/mailroom4.py
```diff
@@ -0,0 +1,251 @@
...
+    sys.exit()  # exit the interactive script
+
+def ck_dir():
+    if not os.path.isdir(DIR_NAME):
+        os.mkdir(DIR_NAME)
+        return False
+    return True
+
+def ss_letterall():
+#     """
```
PythonCHB: you don't want to comment out your docstring!

students/ThanhNhan/lesson06/mailroom4.py
```diff
@@ -0,0 +1,251 @@
...
+#     This method write letter to disk as txt file
+
+#     params:  none
```
PythonCHB: no need to specify None for the params -- you can tell by looking at the function definition.

students/ThanhNhan/lesson06/mailroom4.py
```diff
@@ -0,0 +1,251 @@
...
+#     will check and generate directory
+#     will write all txt file in the directory
+#     """
+
+	#writes each letter to disk as txt file
+	#open ("./thankyou.txt", "w")
+	# tempfile.gettempdir() 
+
+    for key, values in donor_dict.items():
```
PythonCHB: use meaningful names here, too -- yes these are the key and value of the dict, but that's not helpfulu in this context:

`for name, donations in donor_dict.items():`

students/ThanhNhan/lesson06/mailroom4.py
```diff
@@ -0,0 +1,251 @@
...
+    for key, values in donor_dict.items():    
+        letter = create_letter(key, sum(values))
+        filename = key + ".txt"
+        ck_dir()    
+        try:
+            print(filename)
+            filename = os.path.join(DIR_NAME, filename)
+            open(filename, 'w').write(letter)
+        except FileNotFoundError:
+            print("couldn't open missing.txt")
```
PythonCHB: shouldn't this tell the user which dile couldn't be opened? it's probably not called "missing.txt".

And when will you get a FileNotFoundError? In this case, when DIR_NAME is not there.

students/ThanhNhan/lesson06/mailroom4.py
```diff
@@ -0,0 +1,251 @@
...
+        finally:
```
PythonCHB: no need for a finally if it's only pass.

the finally block os for things that need to be done regardless of whether the exeption was raised or not. IF there's nothing to do -- no need to have it.

students/ThanhNhan/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,124 @@
...
+"""
+Send a thank you (adds a new donor or updates existing donor info)
+Create a report
+Send letters (creates files)
+Your unit tests should test the data manipulation logic code: generating thank you text, adding or updating donors, and listing donors.
+Send a thank you (adds a new donor or updates existing donor info)
+"""
+def test_existing_donor():
+
+    new_donor = {"Tom Cruise": [233.23, 12.17],
```
PythonCHB: great to have a dataset for the tests.

But better to have it outside this one test, so you can use it in other tests as well.

students/ThanhNhan/lesson06/mailroom4.py
```diff
@@ -0,0 +1,251 @@
...
+                ln = lookup_name(name) #ln - lookup donor
+                amount = input ("\n Enter amount you want to donate > :")                
+                money_input_ps = ck_dollar(amount)           
+                if (money_input_ps == True):
+                    amount_flt = float(amount)
+                    add_donation(ln, name, amount_flt)
+                    print(create_letter(name, amount_flt))
+        break
+       
+def chk_exist_donor():
```
PythonCHB: strange name for getting all the donors...

students/ThanhNhan/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,124 @@
...
+             "Jessy Test": [33.0],
+             "Jin Unknown": [12.0, 12.0]
+             }
+    mr4.donor_dict = new_donor
+    new_donor_db = mr4.chk_exist_donor()
+    assert new_donor_db == ["Tom Cruise", "Alyssa Nhan", "Jessy Test", "Jin Unknown"]
+
+
+# Validate Name Input
+def test_validate_name():
```
PythonCHB: very nice!

though it's better to make these separate tests -- this way, if the first one fails, the rest won't run.

If this seems tedious -- it is. pytest provides a way to make a bunch of very similar tests liek this, called "parameterized tests" -- you'll learn about that next quarter.

students/ThanhNhan/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,124 @@
...
+    assert name7 is False
+
+    name8 = mr4.validate_name("Tom Asd III")
+    assert name8 is True
+
+    name9 = mr4.validate_name("Tom Asdasfasf, III")
+    assert name9 is True
+
+# Test for lookup Name
+def test_lookup_name():
```
PythonCHB: good!

but again, better as two separate tests.

students/ThanhNhan/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,124 @@
...
+    name1 = mr4.lookup_name("Alyssa Nhan")
+    assert name1 == "Alyssa Nhan"
+
+    name2 = mr4.lookup_name("Thanh Inte")
+    assert name2 is None
+
+#Test dollar amount input
```
PythonCHB: excellent!

students/ThanhNhan/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,124 @@
...
+    donate4 = mr4.ck_dollar("asd")
+    assert donate4 is False
+def test_ck_pt2float():
+    donate4 = mr4.ck_dollar("0.23")
+    assert donate4 is True
+
+#Test for donation add
+def test_existing_donation():
+    mr4.add_donation("Alyssa Nhan", "Alyssa Nhan", 2.32)
+    assert mr4.donor_dict["Alyssa Nhan"] == [23., 23, 2.32]
```
PythonCHB: careful here -- this is assuming a particular state before this test runs. If another test gets added later that adds a donation to this donor, this test will fail, even though it is working correctly.

students/ThanhNhan/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,124 @@
...
+
+def test_new_donation():
+    mr4.add_donation(None,  "Tom Asdf", 1324.23)
+    assert mr4.donor_dict["Tom Asdf"] == [1324.23]
```
PythonCHB: nice!

students/ThanhNhan/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,124 @@
...
+            Dear {},
+            Thank you  for your very kind donation of ${:.2f}
+            It will be put to very good use.
+
+                            Sincerely,
+                            - The Team   
+	        """.format("Tom Tom", 3240.00)
+
+#Check Directorys
+def test_check_directory_default():
```
PythonCHB: this is a tricky one to test.

but rather than expecting a particular state in order to run the tests -- you could have the test code create that dir first, and then check that it's found.

students/ThanhNhan/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,124 @@
...
+    direct = mr4.ck_dir()
+    print("Please make sure directory (D_letter) is create to pass this condition")
+    assert direct is True
+
+#Test Create Report
+# def test_create_report():
+    data = mr4.create_report()
+    print(data)
+    assert data == [["Tom Cruise", 245.40, 2, 122.70],
```
PythonCHB: better not to have your tests rely on a particular test data set, unless it's getting (re-) created in the test. 

the way this is written, if another test adds or removes data, this test will fail.

students/ThanhNhan/lesson06/test_mailroom.py
```diff
@@ -0,0 +1,124 @@
...
+                    ["Jessy Test", 33.00, 1, 33.00],
+                    ["Jin Unknown", 24.00, 2, 12.00],
+                    ["Tom Asdf", 1324.23, 1, 1324.23],]
+ 
+
+#  # Create/save thank you letters for all donors
+def test_save_letters():
+    mr4.ss_letterall()
+    assert os.path.isfile("D_letter/Tom Asdf.txt")
+    assert os.path.isfile("D_letter/Tom Cruise.txt")
```
PythonCHB: modern OSs all support spaces in filenames, but they are still awkward to deal with on the command line and in some code -- so I think it's a good practice to not use spaces in filenames.

PythonCHB:
I've given you credit for all mailrooms except OO mailroom.

Only partial credit for trigrams -- it didn't appear to be done.

If there is anything else you've done that you don't have credit for in Canvas, let me know.

PR #198, opened 2018-12-06
Circle Assignment

thanhnhan7338:
Complete Circle Assignment


PythonCHB:
careful what you put in git. THere is and empty file:

 students/ThanhNhan/lesson04/D

and one called """.py

and all the letters to the donors -- generated files should not be added to version control.

you can do a "git rm the_filename" to remove them

PR #200, opened 2018-12-08
Assignment 06 And Assignment O7 Ready to be graded

timothyjpauley:
Hi, I have updated and these assignment are read to grade

-assignment 06 - 
   MailRoomPart4.py
   test_MailRoomPart4.py all unit tests passed
-assignment 07 -
   html_render.py
   test_html_render.py all unit tests passed

Tomorrow I will turn in 
-assignment08
-assignment09

Thanks, Tim


students/TimPauley/session06/test_MailRoomPart4.py
```diff
@@ -63,7 +64,7 @@ def test_generate_donor_report():
 
     assert report.startswith("Donor Name                | Total Given | Num Gifts | Average Gift")
 
-    assert "Paul Allen                  $    663.23           1   $     663.23" in report
+    assert "Jeff Bezos                  $    877.33           1   $     877.33" in report
```
PythonCHB: you could use a bit more testing here -- did all the donors make it into the report?

and can you test a row in the report without hard coding EXACTLY how it will be laid out?

PythonCHB:
your mailrrom tests are pretty nicely set up to transfer into the OO version -- good work!

PR #201, opened 2018-12-09
added Circle, and modified trigrams to eliminate errors

VioletLuo:


students/ZidanLuo/session08/Circle_Class.py
```diff
@@ -0,0 +1,93 @@
...
+    def __str__(self):
+        return "Circle with radius: {0:.6f}".format(self.radius)
+
+
+    def __repr__(self):
+        return f"Circle({self.radius})"
+
+
+    @classmethod
+    def from_diameter(Circle, value):
```
PythonCHB: we usually use a generic name, like "cls" or "klass" here.

In this case, it is a Circle, but if you subclass, then it will not be anymore -- it could be a Sphere for instance.

students/ZidanLuo/session08/Circle_Class.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def __sub__(self, other):
+        return Circle(self.radius - other.radius)
+
+
+    def __truediv__(self, value):
+        return Circle(self.radius / value)
+
+
+    def __cmp__(self,other):
```
PythonCHB: __cmp__ is generally not used when you have __eq__ and _lt__, etc defined.

In fact, I'm not sure it even gets called at all in Python 3

students/ZidanLuo/session08/Circle_Class.py
```diff
@@ -0,0 +1,93 @@
...
+        return cmp(self.radius, other.radius)
+
+class Sphere(Circle):
+    """
+    There are several ways to initialize the subclass Sphere
+    """
+
+    def __init__(self, radius):
```
PythonCHB: this is exactly the __init__ that's already there -- you don't need to define it at all !

students/ZidanLuo/session08/test_Circle_Class.py
```diff
@@ -0,0 +1,131 @@
...
+    assert (c2 * 3) == Circle(12)
+    assert (3 * c2) == Circle(12)
+
+
+def test_compare():
+    c1 = Circle(2)
+    c2 = Circle(4)
+    c3 = Circle(4)
+
+    assert (c1 > c2) == False
```
PythonCHB: you eantt o use "is" with True, False, and None:

```
assert (c1 > c2) is False
```
etc.

But other wise, really nice tests

students/ZidanLuo/session08/test_Circle_Class.py
```diff
@@ -0,0 +1,131 @@
...
+    assert (c1 < c2) == True
+    assert (c1 == c2) == False
+    assert (c2 == c3) == True
+
```
PythonCHB: you may want to add >=, <= ...

students/ZidanLuo/session08/test_Circle_Class.py
```diff
@@ -0,0 +1,131 @@
...
+
+def test_sort():
```
PythonCHB: good test.

But you defined a sort_key method -- you should test that too.

PR #202, opened 2018-12-10
html render update change and complete

thanhnhan7338:
submit html render


students/ThanhNhan/lesson07/html_render.py
```diff
@@ -0,0 +1,150 @@
...
+        print("kwargs ", self.attributes)
+
+
+    def append(self, new_content):
+        self.contents.append(new_content)
+
+    def render(self, out_file, cur_ind=""):   
+       # self.indent = cur_ind
+        out_file.write(cur_ind + self._open_tag())        
+       # out_file.write("<{}>\n".format(self.tag))
```
PythonCHB: you do want to clean up old code once the new code is working.

students/ThanhNhan/lesson07/html_render.py
```diff
@@ -0,0 +1,150 @@
...
+            try:
+                line.render(out_file, cur_ind + self.indent)
+            except AttributeError:
+                out_file.write(cur_ind + self.indent + line)
+                out_file.write("\n")
+        out_file.write(cur_ind + self._close_tag())
+
+    def _open_tag(self):
+        open_tag = []
+        if (self.attributes is None):
```
PythonCHB: will `self.attributes` ever be None? I think it will always exist -- if not attributes, it will be an empty dict.

PythonCHB: I just removed this line and all tests still passed.

students/ThanhNhan/lesson07/html_render.py
```diff
@@ -0,0 +1,150 @@
...
+            open_tag.append("<{}>".format(self.tag))
+        else:
+            open_tag.append("<{}".format(self.tag))
+            for key, value in self.attributes.items():       
+                open_tag.append(' {}="{}"'.format(key,value))
+            open_tag.append(">\n")
```
PythonCHB: putting the final "\n" here may cause trouble -- probably better to let the render method handle that.

students/ThanhNhan/lesson07/html_render.py
```diff
@@ -0,0 +1,150 @@
...
+
+class OneLineTag(Element):
+
+    def render(self, out_file, cur_ind=""):
+
+        if self.attributes is None:
+            out_file.write(cur_ind + "<{}>".format(self.tag))
+            out_file.write(self.contents[0])
+            out_file.write("</{}>\n".format(self.tag))
+        else:
```
PythonCHB: there is an _open_tag method in the Element class that you can use here: DRY

(don't repeat yourself).

students/ThanhNhan/lesson07/html_render.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+    def append(self, content):      
+        raise NotImplementedError("Cannot add content to OneLineTag")
+
+class Title(OneLineTag):
+    tag = "title"
+
+class H(OneLineTag):
+    tag = "h"
```
PythonCHB: you don't actually need it here -- it will get overridden by self.tag

students/ThanhNhan/lesson07/html_render.py
```diff
@@ -0,0 +1,150 @@
...
+        super().__init__(content, **kwargs)
+        #Element.__init__(self, content=content, **kwargs)
+
+    def append(self, *args):
+        raise TypeError("You can not add content to a SelfClosingTag")
+
+    # loop through the list of contents:
+    def render(self, outfile, cur_ind=""):
+        tag = [cur_ind + self._open_tag()]
+        for key, value in self.attributes.items():
```
PythonCHB: also same code as in the _open_tag method.

students/ThanhNhan/lesson07/html_render.py
```diff
@@ -0,0 +1,150 @@
...
+
+    # loop through the list of contents:
+    def render(self, outfile, cur_ind=""):
+        tag = [cur_ind + self._open_tag()]
+        for key, value in self.attributes.items():       
+            tag.append(' {}="{}"'.format(key,value))
+        tag.append(self._close_tag())
+        outfile.write("".join(tag))
+
+    def _open_tag(self):
```
PythonCHB: is this being used anywhere?

students/ThanhNhan/lesson07/test_html_render.py
```diff
@@ -0,0 +1,459 @@
...
+########
+# Step 1
+########
+
+def test_init():
+    """
+    This only tests that it can be initialized with and without
+    some content -- but it's a start
+    """
+    # e = Element()
```
PythonCHB: non need to comment these out -- they should keep passing!

students/ThanhNhan/lesson07/test_html_render.py
```diff
@@ -0,0 +1,459 @@
...
+
+
+def test_append():
+    """
+    This tests that you can append text
+
+    It doesn't test if it works --
+    that will be covered by the render test later
+    """
+    # e = Element("this is some text")
```
PythonCHB: this too -- might as well keep it -- or remove entire test. An empty test does not good.

students/ThanhNhan/lesson07/test_html_render.py
```diff
@@ -0,0 +1,459 @@
...
+
+
+def test_render_element():
+    """
+    Tests whether the Element can render two pieces of text
+    So it is also testing that the append method works correctly.
+
+    It is not testing whether indentation or line feeds are correct.
+    """
+    # e = Element("this is some text")
```
PythonCHB: and these -- I uncommented all this, and it still runs fine :-)

students/ThanhNhan/lesson07/test_html_render.py
```diff
@@ -0,0 +1,459 @@
...
+    lines = file_contents.split("\n")
+    print(lines)
+    print(lines[1].startswith(Element.indent))
+    assert lines[1].startswith(Element.indent)
+
+
+def test_multiple_indent():   
+#     """
+#     make sure multiple levels get indented fully
+#     """
```
PythonCHB: might as well uncomment the docstring :-)

students/ThanhNhan/lesson07/test_html_render.py
```diff
@@ -0,0 +1,459 @@
...
+    lines = file_contents.split("\n")
+    
+    
+    for i in range(3):  # this needed to be adapted to the <DOCTYPE> tag
+        assert lines[i + 1].startswith(i * Element.indent + "<")
+
+
+    assert lines[4].startswith(3 * Element.indent + "some")   
+
+# def test_element_indent1():
```
PythonCHB: you should uncomment this test -- the content got added to the previous test, but it's better to have separate tests, so if something goes wrong, you know what it it.

PR #204, opened 2018-12-10
Circle Class and other updates

ericstreit:
still working on final mailroom but here is some of the older stuff. Primarily the Circle Class work. Only made it partially through the html renderer

### DIFF
--- a/students/ericstreit/session07/html_render.py
+++ b/students/ericstreit/session07/html_render.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python3
+####stopped at step 4
+"""
+A class-based system for rendering html.
+"""
+
+
+# This is the framework for the base class
+class Element(object):
+
+    tag = "html"
+
+    def __init__(self, content=None, **kwargs):
+        self.contents = [content]
+        # print("contents is:", self.contents)
+
+    def append(self, new_content):
+        self.contents.append(new_content)
+
+    def render(self, out_file):
+        # Loop through the list of contents:
+        out_file.write("<{}>\n".format(self.tag))
+        for content in self.contents:
+            if content is not None:
+                try:
+                    content.render(out_file)
+                except AttributeError:
+                    out_file.write(content)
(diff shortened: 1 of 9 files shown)


ericstreit:
OK, my OO Mailroom is pretty much done. I still plan on include a sorting function so that displayed reports are done so alphabetically (or by total donation maybe) but if you want to review now, please feel free since it's 95% completed and I doubt I'll do any more retooling since I've spent well over 10 hours on this thing! Thanks for your feedback and hopefully my work isn't too ugly to review, I know that it didn't follow the framework as you had suggested. 

I've never done this much coding before. After reviewing my work, I would also appreciate your thoughts on if I would survive the next class or not.

Thanks for your time!

PythonCHB:
Hopefully I'll have a chance ot give this a proper review soon. But for now, as for the next class.

If you think you could do similar assignments to these without too much pain, then you are ready to move on!

PythonCHB:
Gave you credit -- you pass!

But you really want to be able to "get: OO development. html_render is challenging, but it's supposed to be -- it requires a good understanding of subclassing, instance methods vs class attributes, etc....

So do try to finish that one!

ericstreit:
Thank you, Chris. Much appreciated. I probably will take a shot at the next
class, I would really love to learn more on interacting with sql and
non-sql databases.

OO Mailroom was a challenge and it took a lot of time but felt really
rewarding as things started to come together. I know a lot of it is
probably a mess, structure-wise.

I did do the lesson 5 work that wasn't counted but since I got an 85% and
passed I'm not worried about it. With html render I felt I was just beating
my head against a wall, I just couldn't comprehend how some of those
functions were working. Reading code is very hard for me at the moment but
then again, I went from zero coding to where I am now in the last 6 months.
Over winter break I will try and re-tackle that as well as the sparse array
challenge which I didn't even begin (but wanted to!)

Thanks for your help, take care and have a good holiday :)

PR #206, opened 2018-12-11
Mailroom, now with classes, and docstrings

vanulor:
I finished the mailroom, implemented classes, and testing.  This features a dictionary based menu, exception handling, test based development, and error handling.  I included my test as well.  I ended up with 19 tests and they all pass.   I will try to get back to html render this week, but if not please evaluate it as is.  I am about 90% done with it and it passes all tests.


PythonCHB:
That's a lot!

I'll give it all a quick review

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
+import statistics
+
+
+class Donor():
+    """ Class used to contain donor name,donation, and methods involved with a single donor """
+
+    def __init__(self, name, donation=""):
+        """ Takes a name, and optionally a donation list"""
+        self.name = name
+        self.donation = list(donation)
```
PythonCHB: the list constructor makes a list from an iterable.

if donation is an empty string (the default), then this will get you and empty string. but if it's a number or string with a number in it, you will not get what you expect!

```
In [9]: list("")                                                                
Out[9]: []

In [10]: list("500")                                                            
Out[10]: ['5', '0', '0']

In [11]: list(500)                                                              
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-11-76c72db8cca4> in <module>
----> 1 list(500)

TypeError: 'int' object is not iterable
```

PythonCHB: make sure you test creation of a donor with both nothing AND a donation, of it's an option to do either.

If you want to only create empty donors, that's fine, too, but then don't make it a parameter.

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def add_donation(self, *values):
+        """Takes either a single donation, and appends it to the list, or appends a set of donations"""
+        for value in values:
+            self.donation.append(value)
```
PythonCHB: This could be just `self.donation.extend(values)` 

Have you tested with single numbers and multiple numbers?

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def make_average(self, value):
```
PythonCHB: not the best name -- but not bad ;-)

wait! why does this take a value? it should be providing the average of this donors donations --that is:

(and "value" is not a great name for multiple values...)
```
statistics.mean(self.donations)
```
And how about making it a property:

```
@property
average_donation(self):
    return statistics.mean(self.donations)
```

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+        """Averages the donations"""
+        return statistics.mean(value)
+
+    def make_total(self, value):
+        """Totals the donations"""
+        return sum(value)
+
+    def num_donations(self, value):
```
PythonCHB: the whole point of a class is that you keep the data (donation history) in the object with the methods that act on it.

so all these should be using self.donations to get the values for this donor.

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+        """Counts the donations"""
+        return len(value)
+
+    def make_thank_you(self):
+        """ Returns a thank you note for total donations of donor"""
+        return f'Dear {self.name}, Thank you for your donation of ${self.make_total(self.donation):.2f}. These funds help save the migratory butterflies of New South East.Thank you'
+
+    def make_filename(self):
```
PythonCHB: nice!

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+        """Turns a donor name into a filename"""
+        return f'{self.name.replace(" ","_")}.txt'
+
+    def make_report_string(self, value):
+        """Makes a single line of the report(Name,Total Donations,Num of Donatioms, Average Donations)"""
+        return f'{self.name:<20} ${self.make_total(value):<15.2f} {self.num_donations(value):<15} ${self.make_average(value):<15.2f}'
+
+    def __lt__(self, other):
```
PythonCHB: do you need these? maybe ...

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+        return (self.make_total(self.donation))
+
+
+class Donors():
+    """ This contains a list of classes, used to store multiple instances of the Donor class for grouping"""
+
+    def __init__(self, donor_list):
+        self.donor_list = donor_list
+
+    def append(self, cls):
```
PythonCHB: what are you appending? presumable a Donor -- so call it "donor" or something like that.

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+        """Adds another donor to the donor_list"""
+        self.donor_list.append(cls)
+
+    def list_donors(self):
```
PythonCHB: nice!

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def make_headers_string(self):
+        """Header for the report"""
+        headers = ("Donor Names", "Total Given", "Num Gifts", "Average Gifts")
+        header_string = f'{headers[0]:<20} {headers[1]:<15} {headers[2]:<15} {headers[3]:<15}'
+        return header_string
+
+    def sort_donors(self):
+        """Takes a list of donors and sorts by total donations"""
+        self.donor_list = sorted(self.donor_list)
```
PythonCHB: so that's what the comparison operators are for -- nice!

though you defined a sort_key method -- why not use that?

PythonCHB: and if you are going to replace it with the sorted version -- you can use .sort() 👍 

``self.donor_list.sort()``

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+    def make_report(self):
+        """Combines the report header with individual report strings"""
+        report = self.make_headers_string() + "\n"
+        self.sort_donors()
+        for donor in self.donor_list:
+            report += donor.make_report_string(donor.donation) + "\n"
+        return report
+
+    def make_thank_notes(self):
+        """Provides a list containing all the thank you notes for Donors, Not Implemented in final code"""
```
PythonCHB: it looks implemented -- what's missing?

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+        self.thank_notes = []
+        for donor in self.donor_list:
+            self.thank_notes.append(donor.make_thank_you())
+        return self.thank_notes
+
+    def find_donor(self, value):
+        """Takes a string value and returns a donor class if donor name matches string, or a empty list if not"""
+        donor_class = []
+        for donor in self.donor_list:
```
PythonCHB: maybe you want to use a dict for faster lookup, but this is fine too.

It'll get slow if the list gets big -- but then you can refactor -- that's one good reason to put this in a class!

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+            if donor.name.lower() == value.lower():
+                donor_class = donor
+        return donor_class
+
+    def make_filenames(self):
```
PythonCHB: caerful with this -- you are making a full list of filenames separate from the letter list -- they could get out of sync.

and maybe making the filename should be the job of the donor? that one could go either way.

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+        """Makes a list of filenames for the donor list, Not implemented in final code"""
+        filenames = []
+        for donor in self.donor_list:
+            filenames.append(donor.make_filename())
+        return filenames
+
+
+def send_file(msg, filename):
```
PythonCHB: this could be in the Donor class.

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def add_donation(d_list, full_name):
+    """Takes the list of donor classes and a full name and adds a donation using user input"""
+    donor = d_list.find_donor(full_name)
+    if type(donor) == list:
+        donor = Donor(full_name)
+        d_list.append(donor)
+    donation = input("Please enter a donation==>")
+    while not donation.isdigit():
```
PythonCHB: does this work for float values?  i.e. 100.50 ?

the most robust way is:

```
try:
    donation = float(donations)
except ValueError:
    handle error
```

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def send_file(msg, filename):
+    """Take a string and a filename and write the msg out to the file specified"""
+    with open(filename, "w") as outfile:
+        outfile.write(msg)
+
+
+def add_donation(d_list, full_name):
+    """Takes the list of donor classes and a full name and adds a donation using user input"""
```
PythonCHB: Keep your docstring updated -- this is now taking a Donors object, yes?

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+    donation = float(donation)
+    donor.add_donation(donation)
+    send_file(donor.make_thank_you(), donor.make_filename())
+
+
+def thank_you(d_list):
+    """Gnerates a thank you for a new donation, using user input it either lists the donors, or otherwise adds a donation to the list"""
+    while True:
+        full_name = input("Please enter a Full Name==>")
+        if full_name.lower() == "list":
```
PythonCHB: this logic should be in the Donors class -- not here! This code shoulnd't know how the Donors are stored, nor how to find them.

don't you have a method for this in your Donors class?

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+            for name in d_list.list_donors():
+                print(name)
+        else:
+            add_donation(d_list, full_name)
+            break
+
+
+def report(d_list):
+    """Generates the report to the screen"""
+    print(d_list.make_report())
```
PythonCHB: yes!

students/BrianNisonger/session9/mailroom.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def send_donors(d_list):
+    """Writes out a thank you note for each donor in the donor list, amount in file is total donations given"""
+    for donor in d_list.donor_list:
```
PythonCHB: again, at this point, you should not know that the Donor class has a donor_list you can loop through...

students/BrianNisonger/session9/test_mailroom.py
```diff
@@ -0,0 +1,181 @@
...
+
+def test_make_donor():
+    donor = Donor("Fred Flinstone")
+    assert donor.name == "Fred Flinstone"
+
+
+def test_make_donor_with_donation():
+    donor = Donor("Fred Flinstone", [100, 200, 300, 400])
+    assert donor.name == "Fred Flinstone"
+    assert donor.donation == [100, 200, 300, 400]
```
PythonCHB: careful with your naming -- use plural for lists of things -- donations -- and singular for just one -- donation.

students/BrianNisonger/session9/test_mailroom.py
```diff
@@ -0,0 +1,181 @@
...
+        500,
+    ]
+
+
+def test_append_donation():
+    donor = Donor("Fred Flinstone")
+    donor.add_donation(500, 365.00)
+    assert donor.donation == [500, 365.00]
+    donor.add_donation(1007)
+    assert donor.donation == [500, 365.00, 1007]
```
PythonCHB: these tests are good, but do test the implementation a bit too much -- that is, that the Donor object stores the donations in order in a list -- that may not be true in the future.

students/BrianNisonger/session9/test_mailroom.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def test_average():
+    donor = Donor("Fred Flinstone")
+    average = donor.make_average([500, 600, 800])
```
PythonCHB: see my previous notes -- but these should be working with the internally stored donations list.

students/BrianNisonger/session9/test_mailroom.py
```diff
@@ -0,0 +1,181 @@
...
+def test_number_donations():
+    donor = Donor("Fred Flinstone")
+    number_of_donations = donor.num_donations([1, 2, 3, 4, 5])
+    assert number_of_donations == 5
+
+
+def test_thank_you():
+    donor = Donor("Fred Flinstone", [100, 200, 300])
+    thanks = donor.make_thank_you()
+    assert thanks == 'Dear Fred Flinstone, Thank you for your donation of $600.00. These funds help save the migratory butterflies of New South East.Thank you'
```
PythonCHB: with this donation list, the last donation was $300, not $600.

students/BrianNisonger/session9/test_mailroom.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def test_filename():
+    donor = Donor("Fred Flinstone")
+    filename = donor.make_filename()
```
PythonCHB: perfect!

students/BrianNisonger/session9/test_mailroom.py
```diff
@@ -0,0 +1,181 @@
...
+    assert filename == 'Fred_Flinstone.txt'
+
+
+def test_make_report_string():
+    donor = Donor("Fred Flinstone")
+    donation = [100, 200, 300, 400, 500]
+    report_line = donor.make_report_string(donation)
```
PythonCHB: like all the others, this test should be:

```
def test_make_report_string():
    donor = Donor("Fred Flinstone", [100, 200, 300, 400, 500])
    report_line = donor.make_report_string()
    ...
```

students/BrianNisonger/session9/test_mailroom.py
```diff
@@ -0,0 +1,181 @@
...
+        100,
+    ]
+
+
+def test_make_check_Donors_attributes():
+    d1 = Donor("Fred Flinstone", [100, 200, 300, 400])
+    d2 = Donor("James Dean", [500, 600, 700, 800])
+    d3 = Donor("Jack the Ripper", [1, 2, 3, 4])
+    d_list = Donors([d1, d2, d3])
+    assert d_list.donor_list[1].donation == [500, 600, 700, 800]
```
PythonCHB: you don't want to be testing that it's the index-1 in the list -- find it by name, and then test.

That way, you could store it internally differently, and it would all still work.

students/BrianNisonger/session9/test_mailroom.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def test_list_donors():
+    d1 = Donor("Fred Flinstone")
+    d2 = Donor("James Dean")
+    d3 = Donor("Jack the Ripper")
+    d_list = Donors([d1, d2, d3])
+    list_of_donors = d_list.list_donors()
+    assert list_of_donors == [
```
PythonCHB: does list_donors sort the list? if not, then you want to be careful about doing a direct comparison.

students/BrianNisonger/session9/test_mailroom.py
```diff
@@ -0,0 +1,181 @@
...
+    d3 = Donor("Jack the Ripper", [1, 2, 3, 4])
+    d4 = Donor("Mickey Mouse", [100, 200, 300, 5, 15, 7])
+    d_list = Donors([d1, d2, d3, d4])
+    donor_class = d_list.find_donor("Mickey Mouse")
+    assert donor_class.donation == [100, 200, 300, 5, 15, 7]
+    donor_class = d_list.find_donor("")
+    assert donor_class == []
+
+
+def test_make_filenames():
```
PythonCHB: you are already testing the make_filename in the Donor class -- you don't need to test that here, too. -- at least not the names themselves.

PR #208, opened 2018-12-11
Update to miscellaneous files

dvlupr:
Includes many of the missing assignments going back to session 2 or 3.


students/JimJenkins/session07/html_render2
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python3
+
+"""
+author: Jim Jenkins (dvlupr)
+Date: 12/10/2018
+
+Note: I didn't write this the way the original assignment focuses, rather it was written for some level of 
+understanding in OOP.
+"""
```
PythonCHB: the assignment is designed to cover certain aspects of OO subclassing, etc.

I can't really see where you're going here...

students/JimJenkins/session07/html_render.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    tag = 'li'
+
+
+class H(OneLineTag):
+    """
+    subclass for H tag from one line tag
+    """
+    def __init__(self, h_counter, header):
+        self.tag = u'h' + str(h_counter)
```
PythonCHB: that `u'` is for Python2 -- and even with that, you can use:

from __future__ import unicode_literals

students/JimJenkins/session07/html_render.py
```diff
@@ -0,0 +1,163 @@
+#!/usr/bin/env python3
```
PythonCHB: this doesn't look finished.

students/JimJenkins/session08/circle.py
```diff
@@ -0,0 +1,16 @@
...
+    @property
+    def diameter(self):
+        return 2 * self.radius
+
+
+c = Circle(5)
+print(c.radius)
+
+c.radius = 7
+print(c.diameter)
```
PythonCHB: more to do here.

PythonCHB:
I've given you credit for a bit more, here, but don't see much that's complete. You're not at 80%

students/JimJenkins/session09/mailroom.py
```diff
@@ -0,0 +1,192 @@
+#!/usr/bin/env python3
+
+"""
+Author: Jim Jenkins
+Date: 12/10/2018
+"""
+
+import cli_main
+
+if __name__ == '__main__':
```
PythonCHB: and yo don't want an if __name__ block here at all, either -- the top level program will be the cli_main file -- over even a file that does nothing but call that.

PR #215, opened 2018-12-13
Session04-06 Exercises and Mailroom2

fyfynguyen:
**Few questions...**

- _**mailroom2.py (lines  61):**_ I could not figure out how to add to a list that was a value inside a dict. Found solution using set.default() that works but not sure I fully understand it. There was only a small section on set.default() in the reading. 

- _**docstrings:**_ are they required for all functions? Also not sure that I entered the correct return values in my doctrings where used. You used them in most the functions in your solution. Wasn't sure if you were just using them to add commentary for us or if you'd include them otherwise. 

- _**PEP8:**_ no pycodestyle errors!

- **_Github:_** noticed that if I have an open PR, anything I push to my own repo is automatically updated in the PR even if it's not ready for code review. Should I be creating a separate branch for each assignment so that PRs will be isolated from each other? So in my student repo, I'd have a separate branch from master, and I'd open a PR right from that branch. Then merge it back to master after its accepted and everything looks good? Are you able to accept / reject part of a PR? Like what if all my assignments in the pull request were good except for one. 

- _**dict_set_lab with comprehensions:**_ the solution for this exercise was really confusing to me. What I have (lines 38-42) works?

- _**Pytest:**_ I ran the command _$ pytest_ and it picked up all test_.py files. All the tests passed but one file says [52%] and the other [100%]. When I run separately they both say [100%]. What does this mean?

screen shot 2018-12-13 at 3 49 12 pm

**Ignore (not ready for code review yet):**
- mailroom3.py
- mailroom5.py
- test_mailroom5.py
- added to .gitignore


.gitignore
```diff
@@ -1,5 +1,10 @@
 .DS_Store
 .git-completion.bash
+
+# ignore data files
```
PythonCHB: please dont change the main .gitignore - there may be reasons to save some .csv or .txt files  in git.

you can put your own .gitignore with this into your own students dir, and then it will only apply to your files.

students/FyfyNguyen/session04/dict_lab.py
```diff
@@ -0,0 +1,73 @@
...
+
+# Create a frozenset with the letters in 'marathon'
+marathon = frozenset('marathon')
+print(marathon)
+
+# Display the union and intersection of the two sets
+print(python.union(marathon))
+print(python.intersection(marathon))
+print(marathon.union(python))
+print(marathon.intersection(python))
```
PythonCHB: all looks good.

students/FyfyNguyen/session04/file_exercise.py
```diff
@@ -0,0 +1,44 @@
...
+
+# Copy a file from a source to a destination (without using shutil,
+# or the OS copy command)
+with open(source, 'rb') as source_file, open(dest, 'wb') as dest_file:
+    dest_file.write(source_file.read())
+
+# Make it work for any size file: i.e. don't read the entire
+# contents of the file into memory at once, break in chunks
+CHUNKSIZE = 1000
+with open(source, 'rb') as source_file, open(dest, 'wb') as dest_file:
```
PythonCHB: perfect!

students/FyfyNguyen/session04/file_exercise.py
```diff
@@ -0,0 +1,44 @@
...
+    while True:
+        data = source_file.read(CHUNKSIZE)
+        if not data:
+            break
+        dest_file.write(data)
+
+# Write script to generate list of languages students know
```
PythonCHB: nicely done

students/FyfyNguyen/session04/mailroom2.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def add_donation(name_entered):
+    """
+    Ask the user for donation amount and add it to donor_db
+
+    :param: the name of the donor
+    """
+    donation = float(input("Enter donation amount >>> "))
+    donor_db.setdefault(name_entered.title(), []).append(donation)
```
PythonCHB: This is a nice way to either create a new donro if there isn't one, or add to the one that's there if there is -- nice find!

However, in this case, I'd probably be more explicit about whether you are adding a new one or adding to an existing donor.

students/FyfyNguyen/session04/mailroom2.py
```diff
@@ -0,0 +1,200 @@
...
+    Confirm the user wants to add a new donor to donor_db
+
+    :param: the name of the donor
+    """
+    response = input(f"Donor does not exist. Add {name_entered.title()} to "
+                     "database? [y/n?] >>> ")
+    if response.strip() == 'y':
+        add_donor(name_entered)
+    else:
+        main_menu()
```
PythonCHB: you generally don't want to call the main_menu from a function that's called from the main_menu. 

As you've seen, it works. But it's building up a "stack" each time the function is called. If this were a long running system, it would build up a huge stack, and might even crash out when a recursion limit is reached.

PythonCHB:
> docstrings: are they required for all functions? Also not sure that I entered the correct return values in my doctrings where used. You used them in most the functions in your solution. Wasn't sure if you were just using them to add commentary for us or if you'd include them otherwise.

As a rule, yes, every function should have a docstring --even if only one line.

If it's a really simple function, and has a self explanatory name, then maybe not...

PythonCHB:
>Github: noticed that if I have an open PR, anything I push to my own repo is automatically updated in the PR even if it's not ready for code review. Should I be creating a separate branch for each assignment so that PRs will be isolated from each other? So in my student repo, I'd have a separate branch from master, and I'd open a PR right from that branch. Then merge it back to master after its accepted and everything looks good? Are you able to accept / reject part of a PR? Like what if all my assignments in the pull request were good except for one.

yes, that "branching" workflow would be better -- keep things clean and easy to separate. Bu tin this class, many students struggle with even this simplified workflow, so I don't ask folks to do that :-)

students/FyfyNguyen/session05/exceptions_lab.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def safe_input(response):
+    """
+    Return None if user enters ^C or ^D instead of raising
+    exceptions EOFError or KeyboardInterrupt.
+    """
+    try:
+        return input(response)
+    except EOFError:
```
PythonCHB: you can also do:

```
except (EOFError, KeyboardInterrupt):
    ...
```
the other form is for when you do something different with the different exceptions.

PythonCHB:
> Pytest: I ran the command $ pytest and it picked up all test_.py files. All the tests passed but one file says [52%] and the other [100%]. When I run separately they both say [100%]. What does this mean?

when it runs both files, it gives a report of how much it's done at each step -- so after doing one, 52% of all the tests have been run.

This is a lot more useful when there are 100 test files...

students/FyfyNguyen/session06/dict_lab_with_comps.py
```diff
@@ -0,0 +1,42 @@
...
+s3 = {i for i in range(21) if not i % 3}
+s4 = {i for i in range(21) if not i % 4}
+
+# Create sequence that holds any arbitrary divisors
+# Loop through that sequence to build sets - you will end up with
+# a list of sets - one set for each divisor in your sequence
+divisors = [2, 3, 4]
+my_sets = []
+
+for idx in divisors:
```
PythonCHB: when you see this construct:

```
a_list = []
a for loop:
    a_list.append
```
you know that that could be a comprehension. IN this case:

```
my_sets = [{i for i in range(21) if not i % idx} for idx in divisors]
```

But honestly, that's getting a bit too nested and hard to read -- so what you've done is fine :-)

PythonCHB:
I"m going to merge, and then fix the .gitignore.

you should do a:

git pull upstream master

to get that fix.

But there is a lot of missing work here!

PR #217, opened 2018-12-14
Worked with print_grid lesson02

framllat:
Made some changes to the solution print_grid.

Will be adding more changes with other solutions. I had significant job reassignment last month that absorbed almost all of my attention. I have been able to attend class, and submit some assignments. Not all have been complete. I intend to take and have signed up for next quarter. 

Hopefully this isn't an issue. My new role is settling down and I will be able to give the course its due attention.
Thanks
Ron Tallman

### DIFF
--- a/examples/Session09/mailroom.py
+++ b/examples/Session09/mailroom.py
@@ -1,0 +1,102 @@
+# this is the primary function for the program
+
+
+class Donor:
+    def __init__(self, name):
+        self.name = name
+        self.donations = []
+
+    def add_donation(self, donation):
+        self.donations.append(donation)
+
+    @property
+    def num_donations(self):
+        return len(self.donations)
+
+
+
+
+
+
+def get_donors_db():
+    donors = {'Bill Gates': [1000, 2000, 3000],
+              'Rick James': [5000, 5000, 6000, 3000],
+              'James Brown': [4000, 10000, 9000, 12000],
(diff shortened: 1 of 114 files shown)


students/student_framllat/session08/the_circle.py
```diff
@@ -0,0 +1,86 @@
...
+        return self.radius == other.radius
+
+    def __lt__(self, other):
+        return self.radius < other.radius
+
+class Sphere(Circle):
+    """
+    Doing the Spere Creation Dance
+    """
+    def volume(self):
```
PythonCHB: you probably want volume to be a property too.

students/student_framllat/session08/test_the_circle.py
```diff
@@ -0,0 +1,131 @@
...
+
+    assert a_list[0] == Circle(5)
+    assert a_list[3] == Circle(20)
+    assert a_list[0] < a_list[1] < a_list[2] < a_list[3]
+
+def test_sphere_vol():
+    s = Sphere(4)
+
+    print(s.volume())
+    assert s.volume() == 268.082573106329
```
PythonCHB: notice that you have to call `.volume`, but not `.area`, etc... that's because it's not a property.

PythonCHB:
Sounds good. 

You need 80% credit to pass -- you are at about 50% now.

If you can't get it done in the next couple days, an incomplete is an option. Let me know.

PR #218, opened 2018-12-15
Mailroom Project Submission

VioletLuo:
Submitted Cli_main.py, donor_models.py, and test_mailroom_OO.py

### DIFF
--- a/students/ZidanLuo/session07/html_render.py
+++ b/students/ZidanLuo/session07/html_render.py
@@ -1,0 +1,160 @@
+#!/usr/bin/env python3
+
+"""
+A class-based system for rendering html.
+"""
+
+
+# This is the framework for the base class
+class Element:
+    
+    tag = "html"
+    indent = "    "
+    def __init__(self, content=None, **kwargs):
+        self.attributes = kwargs
+        if content is None:
+            self.content = []
+        else:
+            self.content = [content]
+
+
+    def append(self, new_content):
+        self.content.append(new_content)
+
+
+    def render(self, out_file, cur_ind=""):
+        
+        # create open tag markdown <tag_name at1: val1, at2: val2>
+        out_file.write(cur_ind)
+        out_file.write(self._open_tag())
+        out_file.write("\n")
+        # iterate over contents, and call render for each sub-element.
+        # sample output should look like:
+        # <tag1 at_1: val_1, at_2: val_2>
+        #   <tag2 at2_1: val2_1>
+        #   ........
+        #   </tag2>
+        for line in self.content:
+            try:
+                line.render(out_file,cur_ind+self.indent)
+            except AttributeError:
+                out_file.write(cur_ind+self.indent)
+                out_file.write(line)
+                out_file.write("\n")
+        # close supreme tag
+        out_file.write(cur_ind)
+        out_file.write(self._close_tag())
+        out_file.write("\n")
+
+    def _open_tag(self):
+        open_tag = ["<{}".format(self.tag)]
+        for key, value in self.attributes.items():
+            open_tag.append(" "+str(key)+"="+ "\"" + str(value) + "\",")
+        if(open_tag[-1][-1] == ','):
+            open_tag[-1] = open_tag[-1][:-1]
+        open_tag.append(">")
+        return "".join(open_tag)
+
+    def _close_tag(self):
+        return "</{}>".format(self.tag)
+
+
+class Html(Element):
+    tag = "html"
+    def render(self, out_file, cur_ind=""):
+        out_file.write(cur_ind + "<!DOCTYPE html>\n")
+        out_file.write(self._open_tag())
+        out_file.write("\n")
+        for line in self.content:
+            try:
+                line.render(out_file, cur_ind+self.indent)
+            except AttributeError:
+                out_file.write(line)
+                out_file.write("\n")
+        # close supreme tag
+        out_file.write(cur_ind)
+        out_file.write(self._close_tag())
+        out_file.write("\n")
+
+
+class Body(Element):
+    tag = "body"
+
+
+class P(Element):
+    tag = "p"
+
+
+class Head(Element):
+    tag = "head"
+
+class Ul(Element):
+    tag = "ul"
+
+class Li(Element):
+    tag = "li"
+
+class OneLineTag(Element):    
+    def render(self, out_file, cur_ind=""):
+        out_file.write(cur_ind)
+        out_file.write(self._open_tag())
+        out_file.write(self.content[0])
+        out_file.write(self._close_tag())
+        out_file.write("\n")
+
+    def append(self, new_content):
+        raise NotImplementedError
+
+class H(OneLineTag):
+    def __init__(self, level, content=None, **kwargs):
+        self.tag = "h"+str(level)
+        super().__init__(content, **kwargs)
+
+
+class A(OneLineTag):
+    tag = "a"
+
+    def __init__(self, link, content=None, **kwargs):
+        kwargs['href'] = link
+        super().__init__(content, **kwargs)
+
+
+class Title(OneLineTag):
+    tag = "title"
+
+
+class SelfClosingTag(Element):
+    '''
+        Selfclosing tags does not take in contents
+    '''
+    def render(self, out_file, ind=""):
+        out_file.write(ind)
+        tag=self._open_tag()[:-1]+ " />\n"
+        out_file.write(tag)
+    # Raise Exception when a selfclosing tags trying to take in content
+    def append(self, *args):
+        raise TypeError("You can not add content to a SelfClosingTag")
+
+
+    def __init__(self, content=None, **kwargs):
+        if content is not None:
+            raise TypeError("SelfClosingTag can not contain any content")
+        super().__init__(content=content, **kwargs)
+    pass
+
+
+class Meta(SelfClosingTag):
+    tag = "meta"
+
+
+class Hr(SelfClosingTag):
+    tag = "hr"
+
+    # def render(self, out_file):
+    #     out_file.write(self._open_tag())
+    #     out_file.write("\n")
+    #     for c in self.ccontent:
+    #         try
+
+class Br(SelfClosingTag):
+    tag = "br"
--- a/students/ZidanLuo/session08/Circle_Class.py
+++ b/students/ZidanLuo/session08/Circle_Class.py
@@ -1,0 +1,93 @@
+from math import pi
+
+class Circle:
+    def __init__(self, radius):
+        self.radius = radius
+        
+    @property
+    def diameter(self):
+        return self.radius * 2
+
+    @diameter.setter
+    def diameter(self, value):
+        self.radius = value / 2
+
+    @property
+    def area(self):
+        return pi * (self.radius ** 2)
+
+
+    def __str__(self):
+        return "Circle with radius: {0:.6f}".format(self.radius)
+
+
+    def __repr__(self):
+        return f"Circle({self.radius})"
+
+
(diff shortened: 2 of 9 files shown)


VioletLuo:
added html_render project

students/ZidanLuo/session07/html_render.py
```diff
@@ -0,0 +1,160 @@
...
+    tag = "hr"
+
+    # def render(self, out_file):
+    #     out_file.write(self._open_tag())
+    #     out_file.write("\n")
+    #     for c in self.ccontent:
+    #         try
+
+class Br(SelfClosingTag):
+    tag = "br"
```
PythonCHB: this all looks good -- nice job!

students/ZidanLuo/session08/Circle_Class.py
```diff
@@ -0,0 +1,93 @@
...
+    def __cmp__(self,other):
+        return cmp(self.radius, other.radius)
+
+class Sphere(Circle):
+    """
+    There are several ways to initialize the subclass Sphere
+    """
+
+    def __init__(self, radius):
+        super(Sphere, self).__init__(radius)
```
PythonCHB: if all __init__ does is call the superclass __Init__, then you don't need one at all, as that's what will happen if there is no __init__

students/ZidanLuo/session08/Circle_Class.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+    @property
+    def volume(self):
+        return (4 / 3) * (pi * self.radius ** 3)
+
+    @property
+    def area(self):
+        return 4 * pi * (self.radius ** 2)
+
```
PythonCHB: perfect!

students/ZidanLuo/session09/cli_main.py
```diff
@@ -0,0 +1,148 @@
...
+donorDb = Donor_Collection()
+last_donor = ""
+last_donation_amount = 0
+OUT_PATH = "thank_you_letters"
+
+def input_handler():
+    while True:
+        response = input(donor_in_prompt)
+        # create a dictionary for user's selections
+        switch_donorDB_dict = {
```
PythonCHB: ideally, you wold define the "switch dict" outside this function, and then could use this same function with different collections of options.

students/ZidanLuo/session09/cli_main.py
```diff
@@ -0,0 +1,148 @@
...
+def convert_to_std(name):
+    # Exception if the name is invalid
+    lst = name.split(" ")
+    lst2 = [x.capitalize() for x in lst]
+    return ' '.join(lst2)
+
+
+def add_donor(donor_name, donation_amount):
+    donorDb.add_donor(donor_name,donation_amount)
+    global last_donation_amount, last_donor
```
PythonCHB: you REALLY don't want to use globals here! The whole point of making it OO is to encapsulate all the info about a donor in one place.

last_donation_amount should be attainable from teh Donor object -- maybe a `.last_donation` property. an last_donor should be passed around somehow through function calls.

The reason for all that is that is very fragile this way -- say your system gets larger, and you have more than one user at a time processing donations -- then "last_donor" would no longer be clearly defined.

students/ZidanLuo/session09/cli_main.py
```diff
@@ -0,0 +1,148 @@
...
+        if checked_response == "y" or checked_response == "Y":
+            # need check validty for float() function
+            try:
+                amount = input("Please enter a number to indicate the donation amount:\n")
+                if float(amount) < 0:
+                    print("Please enter a positive number")
+                    continue
+                add_donor(convert_to_std(response), float(amount))
+                input_handler()
+            except ValueError:
```
PythonCHB: you want to have the `except ValueError` as close to the call to `float` as you can -- it's always possible that add_donor or input)handler() could end up raise the same exception...

students/ZidanLuo/session09/cli_main.py
```diff
@@ -0,0 +1,148 @@
...
+            continue
+
+
+def prepare_to_run():
+    if not os.path.isdir(OUT_PATH):
+        os.mkdir(OUT_PATH)
+
+
+def pre_save_letter():
+    prepare_to_run()
```
PythonCHB: I called this "prepare_to_run" in case there might be more to do in it someday. So it's OK to expect that that's been done when you get here.

But do run it when the program starts up!

students/ZidanLuo/session09/cli_main.py
```diff
@@ -0,0 +1,148 @@
...
+    donorDb.save_all("thank_you_letters")
+
+
+def send_last():
```
PythonCHB: this should be called when it's needed, passing the donor in, so you don't need the global last_donor.

students/ZidanLuo/session09/cli_main.py
```diff
@@ -0,0 +1,148 @@
...
+        # create a dictionary for user's selections
+        switch_donorDB_dict = {
+            "1": send_last,
+            "2": generate_report,
+            "3": pre_save_letter,
+            "e": main,
+            "E": main,
+        }
+        # choose a function depending on response, if the choice is not in dict, catch exception
+        try:
```
PythonCHB: notice that this code is the same as before -- only the dict is different...

students/ZidanLuo/session09/cli_main.py
```diff
@@ -0,0 +1,148 @@
...
+            err_in = 0
+        # create a dictionary for user's selections
+        switch_main_dict = {
+            "1": input_handler,
+            "2": out_handler,
+            "e": kill_program,
+            "E": kill_program,
+        }
+        # choose a function depending on response, if the choice is not in dict, catch exception
+        try:
```
PythonCHB: and now similar code the third time...

students/ZidanLuo/session09/donor_models.py
```diff
@@ -0,0 +1,193 @@
+#!/usr/bin/env python
+import sys
+import os
+
+class Donor_Collection():
+    def __init__(self):
+        self.donor_dict = {"Fred Jones": Donor("Fred Jones", 500)}
```
PythonCHB: you don't want to hard-code any data in the __init__ -- if you want test data (which you do :-) ) -- add it later with a separate function that you can call in the tests, or for examples.

students/ZidanLuo/session09/donor_models.py
```diff
@@ -0,0 +1,193 @@
...
+        self.sorted_dict = {"Fred Jones": Donor("Fred Jones", 500)}
```
PythonCHB: you never want the same data stored twice -- so either sort on demand, or sort whenever you add or remove a donor.

students/ZidanLuo/session09/donor_models.py
```diff
@@ -0,0 +1,193 @@
...
+
+        self.longest_name = "Fred Jones"
```
PythonCHB: also not a good idea to store this -- compute it when you need it.

students/ZidanLuo/session09/donor_models.py
```diff
@@ -0,0 +1,193 @@
...
+    
+    def copy_constructor(self, sampleDB, longest_name, largest_amount):
+        """
+        A copy constructor take in sample database for testin
+
+        :param: a dictionary collection of initialized donation, longest name in sample database, 
+        largest donation amount in the sample database
+         :return: 
+        """
+        self.donor_dict = sampleDB
```
PythonCHB: Yes! this is how you want to populate it with test data.

students/ZidanLuo/session09/donor_models.py
```diff
@@ -0,0 +1,193 @@
...
+        self.longest_name = longest_name
```
PythonCHB: see pervious notes about this.

students/ZidanLuo/session09/donor_models.py
```diff
@@ -0,0 +1,193 @@
...
+
+        :param: name of the donor and donation amount
+        :return: the new donor database
+        """
+        if donor_name in self.donor_dict:
+            self.donor_dict[donor_name].add_donation(donation_amount)
+        else:
+            self.donor_dict[donor_name] = Donor(donor_name, donation_amount)
+
+        if len(donor_name) > len(self.longest_name):
```
PythonCHB: probably better to compute this when you are making the report -- you may add / remove many donors in between reports...

students/ZidanLuo/session09/donor_models.py
```diff
@@ -0,0 +1,193 @@
...
+            self.longest_name = donor_name
+
+        if self.donor_dict[donor_name].sum_donations > self.largest_amount:
+            self.largest_amount = self.donor_dict[donor_name].sum_donations
+
+    
+    def sort_report(self):
```
PythonCHB: nicely done!

students/ZidanLuo/session09/donor_models.py
```diff
@@ -0,0 +1,193 @@
...
+
+    
+    def send_single(self, donor, last_donation_amount):
+        """
+        find the donor in the dictionary and get the formatted thank you letter
+
+        :param: name of the donor and donation amount
+        :return: a string type formatted thank you letter
+        """
+        return self.donor_dict[donor].letter_generator(last_donation_amount)
```
PythonCHB: nice! -- though shouldn't the donor know it's last donation amount?

students/ZidanLuo/session09/donor_models.py
```diff
@@ -0,0 +1,193 @@
...
+
+        :return: a string type report
+        """
+        report = []
+        width = len(self.longest_name) + 8
+        width2 = len(str(self.largest_amount)) + 8
+
+        self.sort_report()
+
+        report.append("{:{width}}{}{:^{width2}}{}{:^10}{}{:^{width2}}{}\n".format(
```
PythonCHB: nice!

students/ZidanLuo/session09/donor_models.py
```diff
@@ -0,0 +1,193 @@
...
+        """
+        evoke save file function in donor class for each donor
+
+        :param: string of output path
+        :return: 
+        """
+        [donor.save_to_disk(out_path) for donor in self.donor_dict.values()]
+
+    
+    def get_longest_name(self):
```
PythonCHB: NEVER write getters like this in Python! these are all attributes, you can simply access them:

`the_db.longest_name`

Then, if you ever do need to add some logic, you can make them properties.

students/ZidanLuo/session09/donor_models.py
```diff
@@ -0,0 +1,193 @@
...
+        """
+        append new donation to donation list
+
+        :param: donation amount
+        :return: 
+        """
+        self.donations.append(donation)
+
+    
+    def save_to_disk(self, out_path):
```
PythonCHB: nice!

students/ZidanLuo/session09/donor_models.py
```diff
@@ -0,0 +1,193 @@
...
+class Donor_Collection():
+    def __init__(self):
+        self.donor_dict = {"Fred Jones": Donor("Fred Jones", 500)}
+        self.sorted_dict = {"Fred Jones": Donor("Fred Jones", 500)}
+
+        self.longest_name = "Fred Jones"
+        self.largest_amount = 500
+
+    
+    def copy_constructor(self, sampleDB, longest_name, largest_amount):
```
PythonCHB: not a great name -- it's not really copying, it's setting with a bunch of data that are in a different format.

If you do want to make a "real" copy constructor, you can define a __copy__ method, and then the Python copy module will use it.

PR #221, opened 2018-12-16
OO mailroom

Misseswaddles:
Completed OO mailroom with test suite and cli.


students/C_Farris/session09/OO_Mailroom.py
```diff
@@ -0,0 +1,194 @@
...
+        a numeric value.""" 
+        try:
+            value = donation/2
+        except TypeError:
+            return "Donation value must be entirely numeric!"
+        else:
+            if value <=0.005:
+                return "Donation cannot be smaller than a penny!"
+            else:
+                self.donations.append(donation)
```
charlesrobison: Something odd going on here.  When adding a donation, it's overwriting the existing donation, rather than adding to it.

PR #223, opened 2018-12-16
HW [-1] Mailroom_OO

dancornutt:
Got mailroom_oo working. Reads yaml file for db input, uses dictionary with name as key and donor as value. There is always more to improve on but this is a good breaking point. Perhaps I'll work on it over break to improve the user experience and reporting.


PythonCHB:
Lots of versions here -- not sure which to review!

We've given you full credit.

And the yaml is pretty slick -- nice!

PR #226, opened 2018-12-17
Missing Exercises

fyfynguyen:


PythonCHB:
FyFy -- yup, html_render is challenging -- that's why it's so many points. 

Taking a look ....

PythonCHB:
you've accidentally added __pycache__ directories -- be careful!

I think that happened because you deleted the .gitignore.

but also -- you need to be careful with `git add` -- not to add stuff you don't want.

PythonCHB:
giving you full credit to html_render, which brings you JUST 20 80%.

But you really want to do OO mailroom and html_render to make sure you understand OO programming.

I also don't see a trigrams -- that's another good one -- maybe go back and give it a try again -- you may be surprised, now that you have a bit more experience.

PythonCHB:
not merging, due to git issues -- and class is over anyway :-)

PR #227, opened 2018-12-17
Last Assignments

ParastooJ:


PythonCHB:
You've passed without html_render.

But that exercise is really key to understanding subclassing and its implications and uses. I do recommend you give it a good try.

PR #229, opened 2018-12-18
homework mailroom4oo

thanhnhan7338:
I did pass test condition, but when I try to put all together. I got a lot of weird data.  I guess I ran out my time. Will have to continue study more during break.


students/ThanhNhan/lesson09/cli_main.py
```diff
@@ -0,0 +1,102 @@
...
+            donor = Donor(name)
+            ck_name = donor.validate_name()
+            if ck_name is False:
+                print ("Invalid Entry - Please re-enter first and last name")
+            elif ck_name is True:
+                donor_db.add_donor(donor)
+                # amount = input ("\n Enter amount you want to donate > :")                
+                # dollar_amount = donor._ck_dollar(amount) 
+                
+                donor.add_donations(1231)
```
PythonCHB: test code you forgot to remove?

students/ThanhNhan/lesson09/cli_main.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    print("Welcome to mailroom")
+    
+    while True:
+        try:
+            response = input(prompt) 
+            switch_fnc_dict.get(response, default)()
+        except NameError:
+            print ("Must enter number from 1-4 from the menu")
+
```
PythonCHB: this is looking good.

students/ThanhNhan/lesson09/mail5.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def add_donor(self, Donor):
+        if self.find_donor() is True:
+            self.data_collection[Donor.name].append(Donor.donations[0])    
+        else:
+            self.data_collection[Donor.name] = Donor.donations
+
+
+
+class Donor(DonorCollection):
```
PythonCHB: you do not want to subclass this from DonorCollection!

Remember the "is a" relationship can you say "A Donor is a DonorCollection" -- nope. So you don't want to subclass!

In this case, you are writing all the methods you use, so it probably works fine -- but be careful!

students/ThanhNhan/lesson09/mail5.py
```diff
@@ -0,0 +1,197 @@
...
+        Create Donor object
+        param name: the full name of the donor
+        param donations: the donations 
+        """
+
+        self.name = name
+        if donations is None:
+            self.donations = []
+        else:
+            self.donations = [donations]
```
PythonCHB: this will crate a list with donations in it -- which is what you want if there is a single donation passed in. Casreful with naming -- calling it "donations" makes me expect it to be more than one!

students/ThanhNhan/lesson09/mail5.py
```diff
@@ -0,0 +1,197 @@
...
+        print("name is {}".format(self.name))
```
PythonCHB: you don't want prints in here -- unless they are temporary for debugging...

students/ThanhNhan/lesson09/mail5.py
```diff
@@ -0,0 +1,197 @@
...
+        print("donations is", self.donations)
+
+    # @property #outside look like attribute
```
PythonCHB: correct to not do this -- this is a classic method -- you need to pass it data to use it.

students/ThanhNhan/lesson09/mail5.py
```diff
@@ -0,0 +1,197 @@
...
+        Add donation amount of user input
+        """
+        donate_amount = self._ck_dollar(amount)
+        self.donations.append(donate_amount)
+    
+    @property #outside look like attribute
+    def num_donations(self):
+        """
+        measure 
+        param donations: iterable donations
```
PythonCHB: this takes no parameters...

students/ThanhNhan/lesson09/mail5.py
```diff
@@ -0,0 +1,197 @@
...
+    
+    def validate_name(self):
+        """
+        This method is validate user name input
+
+        params:  input user name
+        returns true or false if name is correct input
+        """
+
+        if (self.name == " " or self.name.isdigit()):
```
PythonCHB: you want to do this something like: `if not self.name.strip()` -- that way it will capture any amount of whitespace.

students/ThanhNhan/lesson09/mail5.py
```diff
@@ -0,0 +1,197 @@
...
+            raise ValueError("Can't be negative number")
+            
+        return amount_flt
+
+    
+    def validate_name(self):
+        """
+        This method is validate user name input
+
+        params:  input user name
```
PythonCHB: again: no parameters -- it is acting on self -- none to pass in.

though it would probably be better if you DID pass in the name to be validated -- so you could validate before setting it on the instance. LIke `_ck_dollar`

students/ThanhNhan/lesson09/mail5.py
```diff
@@ -0,0 +1,197 @@
...
+        returns true or false if name is correct input
+        """
+
+        if (self.name == " " or self.name.isdigit()):
+            return False
+
+        str_name = self.name.rstrip()
```
PythonCHB: better to use .strip() -- takes whitespace off both ends.

students/ThanhNhan/lesson09/mail5.py
```diff
@@ -0,0 +1,197 @@
...
+        # #parse first and last name into string
+        try:
+            str_name = str_name.split(' ') 
+            if (str_name[1] == ' '):
```
PythonCHB: I don't think split() will ever return an space for an item. 

YOu want to check how long str_name is -- that will tell you how many parts are in the name.

students/ThanhNhan/lesson09/mail5.py
```diff
@@ -0,0 +1,197 @@
...
+        if (self.name == " " or self.name.isdigit()):
+            return False
+
+        str_name = self.name.rstrip()
+        # #parse first and last name into string
+        try:
+            str_name = str_name.split(' ') 
+            if (str_name[1] == ' '):    
+                return False
+        except IndexError:
```
PythonCHB: this will work -- but it's not checking for ' '.

students/ThanhNhan/lesson09/mail5.py
```diff
@@ -0,0 +1,197 @@
...
+            print ("Need to have first and last name within space between")
+            return False
+
+        #Validate user not to enter digit  
+        for char in str_name:  
+            for i in char:
+                if (i.isnumeric() == True):
```
PythonCHB: can just do:

`if i.isnumeric()`

students/ThanhNhan/lesson09/test_mr_5.py
```diff
@@ -0,0 +1,177 @@
...
+# Count how many time donation
+def test_add_donation():
+    donor = Donor("Fred Test")
+    donor.add_donations(500)
+    assert donor.num_donations == 1
+    donor.add_donations(320)
+    assert donor.num_donations == 2
+
+#Test dollar amount input
+def test_ck_char_input():
```
PythonCHB: nice!

but better to make these separate tests -- so if one fails, the others will still run.

students/ThanhNhan/lesson09/test_mr_5.py
```diff
@@ -0,0 +1,177 @@
...
+        donor.add_donations(-50)
+
+    with pytest.raises(ValueError):
+        donor.add_donations(0.0)
+
+
+# Validate Name Input
+def test_validate_name():
+    name1 = Donor(" ")
+    assert name1.validate_name() is False
```
PythonCHB: or:

`assert not name1.validate_name()`

but this works, too.

students/ThanhNhan/lesson09/test_mr_5.py
```diff
@@ -0,0 +1,177 @@
...
+    assert name9.validate_name() is True
+
+    name10 = Donor("Tom Asdasfasf ")
+    assert name10.validate_name() is True
+
+
+def test_donor_thank_you_Letter():
+    donor = Donor("Tom Tom")
+    donor.add_donations(2432)
+    assert donor.generate_letter(donor.donations) == """
```
PythonCHB: Ideally, you don't test the precise result -- then if it changes, you'll have to change the whole test, too.

students/ThanhNhan/lesson09/test_mr_5.py
```diff
@@ -0,0 +1,177 @@
...
+
+                            Sincerely,
+                            - The Team   
+	        """.format("Tom Tom", 2432.00)
+
+
+
+def test_DonorCollection():
+    donor = DonorCollection(donor_db)
+    assert donor.data_collection == {"Tom Cruise": [233.23, 12.17],
```
PythonCHB: here you have tested internal representation -- better not to do that.

rather,  use a listing of donor names or something to make sure it got stored correctly.

students/ThanhNhan/lesson09/test_mr_5.py
```diff
@@ -0,0 +1,177 @@
...
+                                    "Alyssa Nhan": [23., 23],
+                                    "Jessy Apple, IIS": [33.0],
+                                    "Jin Unknown": [12.0, 12.0]
+                                    }
+    
+def test_exist_list():
+    donor = DonorCollection(donor_db)
+    assert donor.exist_list() == ["Tom Cruise", "Alyssa Nhan", "Jessy Apple, IIS", "Jin Unknown"]
```
PythonCHB: will that list always be in the same order?

students/ThanhNhan/lesson09/test_mr_5.py
```diff
@@ -0,0 +1,177 @@
...
+
+def test_lookup_name():
+    donor = DonorCollection(donor_db, "Test As")
+    assert donor.find_donor() is False
+    donor = DonorCollection(donor_db, "Tom Cruise")
+    assert donor.find_donor() is True
```
PythonCHB: assert is looking for Truthiness already, so just:

assert donor.find_donor() is fine.

But this is a very odd API -- if you have the donor object, why do you need to find it??

I htikn this comes from Donor subclassing DonorCollection -- but you really don't want that.

And DonorCollection needs a way to add a donor one by one.

find_donor() should be a method of DonorCollection:

dc = DonorCollection(donor_db)
dc.add_donor("Tom Cruise")

assert dc.find_donor("Tom Cruise")

PR #230, opened 2018-12-27
Requesting help with test_html_render, html_render, and submitting printGrid

GuyCChan:
Hi Chris,

I have been stuck in the html assignment for quite a while and heading nowhere. Hope you can provide some pointers. I have all intentions of finishing this as I've learned quite a lot while working on it. I'm still trying to finish and submit the ooMailRoom assignment by the end of the year. Thank you for your input.

Guy


PythonCHB:
where are you stuck? any particular line of code, or bug you can't figure out?

students/GuyCChan/session08/circleClass.py
```diff
@@ -0,0 +1,117 @@
...
+    @property
+    def area(self):
+        """ Find area using radius."""
+        return pi * self.radius**2
+
+# classmethod to find area using radius. I still haven't figured out the difference between this and the last"""
+# If this is a classmethod, then the first parameter should be 'cls' instead of 'self'. Please explain."""
+# This should only be necessary if the above method was not included at initialization of the class but is """
+# added later. Is this correct? """
+#     @classmethod
```
PythonCHB, line 32: The point of the "from_diameter" classmethod is to provide a way to create a new circle when you have the diameter. You don't need a "from_radius" because the regular initializer already uses the radius to make the Circle.

students/GuyCChan/session08/circleClass.py
```diff
@@ -0,0 +1,117 @@
...
+#     def area_from_rad(cls):
+#         return cls(self.radius)**2 * pi
+
+    @property
+    def area_from_diam(self):
```
PythonCHB, line 37: There is no need to compute the area from the diameter -- if you want to know the area of the circle, it get computed from the radius.

students/GuyCChan/session08/circleClass.py
```diff
@@ -0,0 +1,117 @@
...
+    def __str__(self):
+        """ print the informal representation of an object, or string."""
+        return f"A circle with radius: {self.radius}"
+
+    @property
+    def __repr__(self):
+        """ Print the formal representation of an object with more precise values when appropriate."""
+        return f"Circle ({self.radius})"
+
+    def add_radius(self, other):
```
PythonCHB, line 56: this should be the __add__ method -- it adds two circles together.

PythonCHB (approved):
I think we can call Circle done :-)

students/GuyCChan/session02/printGrid.py
```diff
@@ -0,0 +1,51 @@
...
+# Param m specifies the number of rows and columns and param n, the cell size.
+def print_grid_2_param(m,n):
+    line = ("+ " + "- " * n) * m + "+"
+    stik = ("| " + "  " * n) * m + "|"
+    print((line + "\n" + (stik + "\n") * n) * m + line)
+
+
+print_grid_2_param(3,4)
+
+
```
PythonCHB, line 51: this looks good!

students/GuyCChan/session07/html_render.py
```diff
@@ -0,0 +1,69 @@
...
+
+class Head(Element):
+    tag = "head"
+
+
+class OneLineTag(Element):
+    pass
+
+    def render(self, out_file, **attributes):
+        open_tag.write("<{} ".format(self.tag))
```
PythonCHB: just ran the tests -- one failiure is on this line:

```
self = <html_render.Title object at 0x109248a90>
out_file = <_io.StringIO object at 0x1092639d8>, attributes = {}

    def render(self, out_file, **attributes):
>       open_tag.write("<{} ".format(self.tag))
E       NameError: name 'open_tag' is not defined

html_render.py:54: NameError
```

well, yes, there is no "open_tag" defined. and it there were, it would be a method or a string, and would not hvea a write method.

What you need to do is define an open_tag method that generates the open tag with the attributes. then you could use that here:

out_file.write(self.open_tag)

And all elements will need that open_tag with attributes, so it should go in the Element base class.

PythonCHB: THe otyher failure is here:

```
_______________________________ test_attributes ________________________________

    def test_attributes():
        e = P("A paragraph of text", style="text-align: center", id="intro")
        file_contents = render_result(e).strip()
        print(file_contents)
        assert "A paragraph of text" in file_contents
        assert file_contents.endswith("</p>")
>       assert file_contents.startswith("<p ")
E       AssertionError: assert False
E        +  where False = <built-in method startswith of str object at 0x109274940>('<p ')
E        +    where <built-in method startswith of str object at 0x109274940> = '<p>\nA paragraph of text\n</p>'.startswith

test_html_render.py:210: AssertionError
----------------------------- Captured stdout call -----------------------------
<p>
A paragraph of text
</p>
```

This test is testing if the attributes are rendered properly for a paragraph element (that is, any element, really)

So work on this first -- this requires adding the attributes to the opening tag, probably in an open_tag method, and using it in the main Element render method.

PythonCHB (commented):
getting closer...

students/GuyCChan/session09/test_objOriMailRoom.py
```diff
@@ -0,0 +1,226 @@
+from objOriMailRoom import Donor, DonorCollection, make_report, letters_to_donors, test_data
+import pytest
+
+# Tests for Donor class.
+def test_init():
+    db = donors_db()
+    assert len(donors_db) == 0
+
+def test_Donor_in_db():
```
PythonCHB: make your test names meaningful - this is not testing if there is a donor on the db...

students/GuyCChan/session09/test_objOriMailRoom.py
```diff
@@ -0,0 +1,226 @@
...
+    donor = donor("Mike May")
+    assert donor.name == "Mike May"
+    assert donor.donation is None
```
PythonCHB: a Donor may have had multiple donations -- so what does donor.donation mean? the last one? all of them, ? 

having it be None at this stage is fine, but what it the end goal for this attribute (or property)?

students/GuyCChan/session09/test_objOriMailRoom.py
```diff
@@ -0,0 +1,226 @@
...
+
+def test_add_donation():
+    add_donation = self.donations("100.00")
```
PythonCHB: Donor.donations is a method? that takes a string? probably not what you want. and there is not self defined here...

this should probably look something like:

```
def test_add_donation():
    donor = Donor("Karl Marx")
    donor.add_donation(100)
    assert donor.last_donation == 100
```

students/GuyCChan/session09/test_objOriMailRoom.py
```diff
@@ -0,0 +1,226 @@
...
+    assert add_donation == 100.00
+
+def test_tot_donations():
```
PythonCHB, line 35: this should be somethign like:

```
def test_tot_donations():
    donor = Donor("Teri Thomas":, [60.00, 49,50, 100.50])
   assert donor.total_donations = 210.0
```

students/GuyCChan/session09/test_objOriMailRoom.py
```diff
@@ -0,0 +1,226 @@
...
+    db = donors_db({"Teri Thomas": [60.00, 49,50, 100.50]})
+    assert sum(self.donations) == 210.00
```
PythonCHB: there is no "self" here -- self is only for within method definitions.

"self" means "the current instance of the class" -- when you are using an instance of the class, then it it referred to by whatever name you gave it.

PythonCHB (approved):
run your code (idally the tests) often -- you should never have this much non-working code at once.

work on one small piece at at time: decide what it needs to do, write a test that confims it is doing that, and then write the code until the test passes. One then should you move on the the next bit of functionaility.

And as you design your classes, be clear on what they are supposed to be used fo:

what data does that have? -- those are attributes, or maybe properties
what do they need to do? -- those are methods.

Example:

A Donor has a name and a collection of past donations. So it will probably have attributes like:

self.name
self.donations

You also probably need to know what the last doantion was: that can can be a property:

```
@property 
def last_donation
    return self.donations[-1]
```
and maybe total_donations, similarly....

students/GuyCChan/session09/test_objOriMailRoom.py
```diff
@@ -1,226 +1,73 @@
...
-accesses information about a single donor, then it should most likely live in this class.
-
-DonorCollection Class
-Class responsible for donor collection data encapsulation
-
-This class will hold all of the donor objects, as well as methods to add a new donor, search for a given donor,
-etc. If you want a way to save and re-load your data, this class would hold that method, too.
-
-Your class for the collection of donors will also hold the code that generates reports about multiple donors.
+def test_thank_all_donors():
```
PythonCHB, line 66: this file won't run -- I get:

```
E     File "/Users/Chris/PythonStuff/UWPCE/Fall2018-PY210A/notes/GuysRepo/students/GuyCChan/session09/test_objOriMailRoom.py", line 66
E       def test_thank_all_donors():
E         ^
E   IndentationError: expected an indented block
```

which makes sense -- you have a bunch of def lines with no function bodies -- those should be commented out, or you can put a "pass" in each to make it syntactically correct. Though that will show those tests as passing, which might be a bit deceptive :-)

students/GuyCChan/session09/test_objOriMailRoom.py
```diff
@@ -1,226 +1,73 @@
-from objOriMailRoom import Donor, DonorCollection, make_report, letters_to_donors, test_data
+from objOriMailRoom import Donor, test_data, DonorCollection # make_report, letters_to_donors, test_data
 import pytest
 
-# Tests for Donor class.
-def test_init():
-    db = donors_db()
-    assert len(donors_db) == 0
+sample_db = DonorCollection(test_data())
```
PythonCHB, line 4: fixing the sytax error to get the tests to run, I get this now:

```
==================================== ERRORS ====================================
___________________ ERROR collecting test_objOriMailRoom.py ____________________
test_objOriMailRoom.py:4: in <module>
    sample_db = DonorCollection(test_data())
objOriMailRoom.py:66: in __init__
    self.donor_info.update({donor.name: donor})
E   AttributeError: 'DonorCollection' object has no attribute 'donor_info'
!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!
=========================== 1 error in 0.16 seconds ============================
```
I'll comment on that code in place.

students/GuyCChan/session09/objOriMailRoom.py
```diff
@@ -51,44 +51,42 @@ def avg_donation(self):
...
     def __init__(self, donors=None):
-        """ Initialize the DonorCollection class to hold info, data, and methods for all donors in database."""
+        """ Initialize the DonorCollection class to hold donors info and methods for donors in database."""
         if donors is None:
-            self.donor_inf = {}
+            self.donor_info = {}
         else:
-            self.donor_inf = {name: donations}
+            for donor in donors:
+                self.donor_info.update({donor.name: donor})
```
PythonCHB, line 66: ``self.donor_info`` does not yet exist -- you need to create an empty dict first, and then you can update it.

and this would work if it did exist, but ``update`` is for updating a dict with the contents of antoher dict -- usally multiple entries -- in this case, it's cleaner to simply do:

```
for donor in donors:
    self.donor_info[donor.name] = donor
```

students/GuyCChan/session09/test_objOriMailRoom.py
```diff
@@ -1,226 +1,73 @@
...
 def test_donors_in_db():
-    assert "John Goodfellow" in test_data
-    assert test_data("Mike May") is None
-    assert len(test_data) == 6
+    assert "John Goodfellow" in sample_db
+    assert sample_db("Mike May") is None
+    assert len(sample_db) == 6
 
 def test_add_donor():
     new_donor = donor("Alison Ailey")
```
PythonCHB, line 55: this gives an error -- the class is ``Donor`` with a capital D.

there are a few more of these in there...

students/GuyCChan/session09/test_objOriMailRoom.py
```diff
@@ -1,226 +1,73 @@
...
+    assert donor[1][-1] == 200.00
+
 
-# Tests for DonorCollection class.
+""" Tests for DonorCollection class."""
 def test_donors_in_db():
-    assert "John Goodfellow" in test_data
-    assert test_data("Mike May") is None
-    assert len(test_data) == 6
+    assert "John Goodfellow" in sample_db
```
PythonCHB, line 50: using ``in`` will only work if you define a __contains__ method in your DonorCollection.

which would be pretty nifty, but having a ``find_donor()`` or ``get_donor()`` method is probably more useful, as, in general, you want not only to know if the donor is there, but then get the donor object if it is there.

students/GuyCChan/session09/test_objOriMailRoom.py
```diff
@@ -1,226 +1,73 @@
...
 
 def test_add_donor():
     new_donor = donor("Alison Ailey")
-    assert "Alison Ailey" in test_data
+    assert "Alison Ailey" in sample_db
     assert new_donor.value() is None
 
 def test_find_donor():
     donor = donor("Dave Day")
     assert test_data(donor) is None
```
PythonCHB, line 61: what is the supposed to be testing???

students/GuyCChan/session07/html_render.py
```diff
@@ -18,22 +19,46 @@ def __init__(self, content=None, **kwargs):
...
 
-    def render(self, out_file):
-        # loop through the list of contents (see func above)
-        out_file.write("<{}>\n".format(self.tag))
+    def _open_close_tags(self):
+        """ Made open and close tags for all elements."""
+        attrs = " ".join(["{}='{}'".format(key, value) for key, value in self.attributes.items()])
+        # I really couldn't figure out comletely how to word these lines, particularly the .strip(),
+        # and had to resort to plagiarising! But now I think I begin to appreciate how it's done.
+        if attrs.strip():
```
PythonCHB, line 27: this could be a touch cleaner with:

```
        if self.attributes:
            attrs = " ".join(["{}='{}'".format(key, value) for key, value in self.attributes.items()])
            open_tag = "<{} {}>".format(self.tag, attrs)
        else:
            open_tag = "<{}>".format(self.tag)
```

but this works, too.

students/GuyCChan/session07/html_render.py
```diff
@@ -18,22 +19,46 @@ def __init__(self, content=None, **kwargs):
...
+        out_file.write(close_tag)
+        out_file.write("\n")
 
 class Html(Element):
     tag = "html"
 
+    def render(self, out_file, cur_ind="", **kwargs):
+        out_file.write(cur_ind)
+        out_file.write("<!DOCTYPE html>")
+        out_file.write("\n")
```
PythonCHB, line 59: or just:

`out_file.write("<!DOCTYPE html>\n")`

:-)

students/GuyCChan/session07/html_render.py
```diff
@@ -50,20 +75,79 @@ class Head(Element):
 class OneLineTag(Element):
     pass
```
PythonCHB, line 76: don't need the pass here, now that there is a method defined.

students/GuyCChan/session07/html_render.py
```diff
@@ -50,20 +75,79 @@ class Head(Element):
...
+            raise TypeError("SelfClosingTag cannot contain any content.")
+        super().__init__(content=content, **kwargs)
+
+
+    def render(self, out_file, cur_ind="", **kwargs):
+        # Please explain the use of the underscore in the following line. I tried and tried for hours
+        # and over 2 days to get this going but failed. Is it the so called "throwaway" variable?
+        # How does it work here? Without it, tests for "hr" would always come back saying "hr" has no
+        # attribute 'open_tag'.
+        open_tag, _ = self._open_close_tags()
```
PythonCHB, line 108: the underscore is a "throwaway" variable -- you could give it any name.

`self._open_close_tags()` returns two values: the opening tag and the closing tag. So you need to assign it to two names, in this case, `open_tag` and `_`. 

If you assign it to one name, then it doesn't raise an error, but rather sets that one name to a tuple with both values in it, so you now have a tuple of two strings, rather than a string.

Technically, functions always return one value -- either a single value or a tuple with multiple values in it. Python's "unpacking" syntax allows you to unpack that tuple inot multiple variables.

So this code could be:
```
tags = self._open_close_tags()
open_tag = tags[0]
```

and get the same results.

students/GuyCChan/session07/html_render.py
```diff
@@ -50,20 +75,79 @@ class Head(Element):
...
+class Ul(Element):
+    tag = "ul"
+
+
+class H(OneLineTag):
+    tag = "h"
+
+    def __init__(self, level, *args, **kwargs):
+        # Please explain why the param "content" has to be replaced with *args
+        # in order to have it work. Why positional arguments?
```
PythonCHB, line 146: you could use "content" here instead, in both places:

```
    def __init__(self, level, content=None, **kwargs):
        self.tag = "h" + str(level)
        super().__init__(content, **kwargs)
```
but you need the `=None` as it's an optional argument.

but the *args, **kwargs approach is a way to be clear that you are passing on exactly what is passed in to the superclass' __init__, without changes. It makes it clear that the API is the same except for the one added "level" parameter.

It also makes it more robust in the face of changes in the superclass -- if we changed the __init__ in the superclass to require one, two (or zero) positional parameters, then *args would still work.

PythonCHB (approved):
Giving full credit for html_render -- nice job -- that's a complex one!

I hope you have a bit more of a handle on subclassing, and method overriding and all that.

mailroom is disapointing -- you have a good start, but none of teh code runs, and there are a LOT of things that don't work (or even run). You should never have that much non-running code in there.

If you are doing a major refactor, like here, I can see how you'd get there, but you want to comment out any tests (or other code) that has not been refactored yet, so you can run what has been updated. So same as when starting from scratch:

1) decide what feature to add or fix
2)write a test for it
3) run the tests, to make sure the test itself runs (expecting it to fail the first time)
4) write some code to implement that feature
5) run the tests
6) if it still fails, go back to (4)

This way, the code will only not run for a short time, while you fix one error, and you know that any error you get (if it's a bunch!) were caused by the last code you changed, which should only be a few lines, so easier to debug.

I hope you'll be able to keep working on it -- but your next step should be to get the tests to all run -- even if that means commenting out most of them.

PythonCHB:
Not merging, as there is some git ugliness with other files that we don't want to change.

And it's the end of class anyway.

GuyCChan:
Thank you. Got your comments. Enough to chew on for a while. Hind sight tells me that you're right: it takes more time and effort to troubleshoot and take out the bugs than to do it the other way around. Thanks again.


Guy


________________________________

